PR #61, opened 2020-07-24
feat: Variant tested inline single and double MPU (mainstandard only)

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-3132

Waiting to hear back from Lisa Townson regarding the addition of the two new ad slots with their respective size allocations to test again. See [this thread in Slack](https://newsuktechnology.slack.com/archives/C018BTNDQ5S/p1596446901004200).

* Created a new `variant-testing` component which reads a `variants` object passed in to RN by native containing a key:value pair for any currently running variant tests and the resulting group i.e. `{ articleMpuTestVariant: "C" }`. Config can then be created for each test based on the group assigned. This has been done for the `articleMpu` test. This config is then shared around the app via context which can be accessed via the `useVariantTestingContext` hook.

* Changed the Ad component to accept a new `width` prop. Previously Ad sizes were determined by screen width only. However, to achieve inline ads with text wrap we need to provide the actual advert width. I have also added two new `slotName`s ("native-inline-ad-b", "native-inline-ad-c") to allow us to set specific advert sizes for each test group assignment (If we have multiple sizes on one slot, we have no control over which size will be displayed as this is determined by winning bid).

* Change the `article-skeleton` `body-utils` function `setAdPosition` to `setupAd`. It is within this function that, depending on the test group assignment, we either return the content unchanged (control group A) or add attributes to the `ad` slot to specify the provided width, height, adPosition, set the `display` to `inline`, and override the `slotName` with the relevant one for that test group. These attributes are then read by the above `Ad` component to modify its behaviour.

* Added a new `isInlineAd` prop to `InlineParagraph` component to determine if the inline component to be rendered is an instance of the `Ad` component. This is also where, with the help of the above context, we can specify the width and height of the `BoxExclusion` which effectively reserves the correct space for the advert.

* Added/Updated tests.

* Removed all references to the previous POC `adPosition`.

Screenshot 2020-08-03 at 18 20 44

Screenshot 2020-08-03 at 18 23 46


packages/article-skeleton/src/article-body/inline-paragraph.js
```diff
@@ -60,6 +63,17 @@ const InlineParagraph = ({
 
   const positioned = manager.layout();
 
+  const getInlineWidthAndHeight = () => {
+    const { articleMpu } = variants;
+
+    if (!isInlineAd || !articleMpu) return { width: contentWidth * 0.35 };
```
aronshamash, line 69: do you think this should return `height: undefined` ?

packages/article-skeleton/src/body-utils.js
```diff
@@ -50,13 +51,30 @@ export const setAdPosition = (adPosition, content) => {
     return !isItemAd;
   });
 
-  if (!currentAdSlotIndex || adPosition === currentAdSlotIndex + 1)
+  if (!currentAdSlotIndex) return content;
+
+  // Article MPU Test
+  const { articleMpu } = variants;
+
+  if (
```
aronshamash, line 59: maybe could be abstracted to a named function to make clearer why this is happening?

packages/variant-testing/src/variant-testing.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { createContext, useContext, ReactNode } from "react";
```
aronshamash, line 1: component looks good

packages/variant-testing/src/variant-testing.tsx
```diff
@@ -0,0 +1,49 @@
...
+      articleMpu: {
+        group: articleMpuTestVariant,
+        ...(articleMpuTestVariant !== "A" && {
+          slotName:
+            articleMpuTestVariant === "B"
+              ? "native-inline-ad-b"
+              : "native-inline-ad-c",
+          adPosition: 5,
+          width: 300,
+          height: articleMpuTestVariant === "B" ? 250 : 600,
```
aronshamash, line 28: why do we need sizes set [here](packages/ad/src/utils/sizes.js) and in this variant config?

packages/article-skeleton/src/article-body/inline-paragraph.js
Geoff-Ford: I did think about that because of the name I gave the function. But it isn't necessary as height is defined or overridden later anyway. I might change the function to `getInlineDimensions`

packages/variant-testing/src/variant-testing.tsx
Geoff-Ford: Well initially I wasn't dynamically setting the `slotName` which now does associate a test group with a set of dimensions. However the sizes are also used to set the `BlockExclusion` dimensions which are outside of the Ad component and its normal remit.

aronshamash: ok thanks. that makes sense.

packages/article-skeleton/src/article-body/inline-paragraph.js
Geoff-Ford: Done

packages/article-skeleton/src/body-utils.js
Geoff-Ford: I have extracted the ad transformation logic out to a separate function `setupArticleMpuTestAd`. But I have left the conditions in the main function as that makes sense to me and also means I don't have to pass several more things into the new function.

packages/ad/src/utils/ad-init.js
```diff
@@ -265,7 +265,7 @@ export default ({ el, data, platform, eventCallback, window }) => {
             if (
               event &&
               event.slot &&
-              event.slot.getSlotElementId() === "native-inline-ad" &&
+              /^native-inline-ad/.test(event.slot.getSlotElementId()) &&
```
c-j-j, line 268: Might be worth leaving a comment here suggesting this is required for a/b testing? Someone might change it back thinking they're not changing behaviour

packages/article-skeleton/src/article-body/article-body-paragraph.js
```diff
@@ -59,6 +60,7 @@ const ArticleParagraph = ({
       str={str}
       scale={scale}
       inline={inline}
+      isInlineAd={inline?.type === Ad}
```
c-j-j, line 63: Is this typescript syntax inside a `.js` file?

Geoff-Ford: Optional chaining is at stage 4 and so is now enabled by default on @babel/present-env

PR #73, opened 2020-07-30
feat: front lead 2 and 2

c-j-j:
Creating draft PR for visibility/early feedback.

The plan for this PR is to have the lead 2 and 2 front-page working for all breakpoints and on landscape/portrait.

### Oustanding issues (later for a PR)
- [ ] Figure out better solution for how to have correct teaser font size on landscape/portrait and medium/wide/huge - do we need to use the font size logic in styleguide (which doesn't yet support such a thing)
- [ ] Byline has incorrect marginBottom
- [ ] Small breakpoint is broken - but as it isn't going to show a front page I'm not sure if this is an issue?
- [ ] Not enough room to display full headlines on an iPad mini
- [ ] Text generally doesn't align to baseline. RemoveWhitespace seems to be overly cautious with how much space to allocate.

### Orientation/Breakpoint limitation

Currently, we are using breakpoints medium/wide/huge. These are incompatible with the designs we have been given as they have differences for landscape/portrait. For example, Tile-X's headline should have 45px on landscape and 50px on portrait when the `wide` breakpoint is hit... so I've had to assign different font sizes based on orientation as well as breakpoint.

There's also an issue with the existing slices. Lead 2 and 2 no pic - for example, only uses breakpoints to decide which layout to show... which results in the landscape template being shown on an iPad Pro 12.9 in portrait because it hits that breakpoint.

Also, in the styleguide package, there is font sizing logic that uses the font-size passed in from native, however, I don't think this is currently used for article teaser text to show larger fonts when the `medium`/`large`/`xlarge` options are set.

Potential solution: In the Responsive context provider, we also provide a `deviceSize` which can be small/medium/large etc, and then use that value to decide how large text should be. We then will only need to factor in edge cases when landscape/portrait have different font sizes (for e.g. tile E has different headline font size)

### New Components Created/Copied 

**Copied Tile X/Y/D/E/AL to Tile-X-Front/etc** 
The front tiles will have the following differences:
- Always show byline, and will be before the content
- Will not need save star/labels/flags
- Will use the new `FrontTileSummary` component
- Have justified/left-aligned text based on template
- Will eventually have columns
- Subtle styling differences - for e.g. different font sizes for landscape/portrait/
- Uses article content as teaser text, instead of the summary800 fields, as some tiles will need a lot of teaser text.

**Copied TileSummary to FrontTileSummary**
- Uses new `FrontArticleSummaryContent` to render the content
- Uses new `FrontArticleSummary`
- Does not render flags/save star

**New FrontArticleSummaryContent**
This does the following, and wraps the original `ArticleSummaryContent` component
- indents paragraphs
- different font styles for landscape/portrait - which we might want to duplicate in `ArticleSummaryContent`

iPad mini in portrait
image

iPad mini in landscape
image


packages/edition-slices/src/tiles/tile-x-front/styles/index.js
```diff
@@ -0,0 +1,50 @@
...
+const fontSizePortraitResolver = {
+  [editionBreakpoints.medium]: 40,
+  [editionBreakpoints.wide]: 50,
+  [editionBreakpoints.huge]: 50,
+};
+
+export default (breakpoint) => ({
+  container: {
+    flex: 1,
+    padding: spacing(2),
```
aronshamash, line 24: can just be `paddingRight: spacing(2), paddingBottom: spacing(2)` I think.

packages/edition-slices/src/tiles/tile-x-front/styles/index.js
```diff
@@ -0,0 +1,50 @@
+import {
+  colours,
+  fonts,
+  spacing,
+  editionBreakpoints,
+  globalSpacingStyles,
+} from "@times-components-native/styleguide";
+
+const fontSizeLandscapeResolver = {
```
aronshamash, line 9: what about combining this to 1 object for both landscape and portrait ?

packages/front-page/front-article-summary-content/styles.js
```diff
@@ -0,0 +1,48 @@
+import styleguide, {
+  editionBreakpoints,
+} from "@times-components-native/styleguide";
+
+const { colours, fonts, spacing } = styleguide();
+
+const fontSizeLandscapeResolver = {
```
aronshamash, line 7: these 4 objects can be combined I think.

packages/slice-layout/src/templates/frontleadtwoandtwo/index.js
```diff
@@ -0,0 +1,111 @@
+import React from "react";
+import { Dimensions, View } from "react-native";
+import { editionBreakpoints } from "@times-components-native/styleguide";
+import { ItemColSeparator } from "../shared";
+import stylesFactory from "./styles";
+
+import VerticalLayout from "../verticallayout";
+
+function calculateContentWidth(windowWidth) {
```
aronshamash, line 9: could this be done in an object literal to reduce complexity ?

packages/edition-slices/src/tiles/tile-x-front/styles/index.js
c-j-j: Unsure how that'll work, as we access these fields with the breakpoint and they're different for landscape/portrait:

```
fontSizeLandscapeResolver[breakpoint]
```

packages/edition-slices/src/tiles/tile-x-front/styles/index.js
c-j-j: Yep that's nicer.

packages/slice-layout/src/templates/frontleadtwoandtwo/index.js
c-j-j: Happy to change this... How would that work with an object? 

I thought this needed to be done via if statements, as the order of the checks is important. For example, we first need to check the widest width and then check the narrower widths.

aronshamash: hmm yes looking at it closer it makes sense to keep as you have it!

packages/edition-slices/src/tiles/tile-x-front/styles/index.js
aronshamash:
```
const fontSizeResolver = {
  [editionBreakpoints.medium]: {
   landscape: 40,
   portrait: 40,
   },
  [editionBreakpoints.wide]: {
   landscape: 40,
   portrait: 50,
   },
  [editionBreakpoints.huge]: {
   landscape: 50,
   portrait: 50,
   }
}
```
than access by `fontSizeResolver[breakpoint].landscape`

c-j-j: Yeah that sounds good - makes it easier to see the differences between landscape and portrait

packages/front-page/front-article-summary-content/styles.js
c-j-j: 2 objects might work better for this, one for font size and one for line height. Similar to https://github.com/newsuk/times-components-native/pull/73#discussion_r466454197 - what do you think?

.eslintrc.js
```diff
@@ -85,6 +85,7 @@ module.exports = {
     "@typescript-eslint/no-inferrable-types": "off",
     "@typescript-eslint/no-non-null-assertion": "off",
     // Typescript specific to revist when majority TS codebase
+    "@typescript-eslint/ban-ts-comment": "off",
```
Geoff-Ford, line 88: Can you bump this above the comment as I can't see us never needing to have some ts comments

packages/edition-slices/src/slices/frontleadtwoandtwo/index.js
```diff
@@ -0,0 +1,99 @@
...
+  // const {
+  //   onPress,
+  // slice: { lead1, lead2, support1, support2 },
+  // } = props;
+
+  const onPress = () => {
+    console.log("PRESSED");
+  };
+
+  const { lead1, lead2, support1, support2 } = testSlice;
```
Geoff-Ford, line 23: Is this `onPress`, the above `testSlice` to be removed in favour of the commented code above?

packages/edition-slices/src/slices/frontleadtwoandtwo/index.js
```diff
@@ -0,0 +1,99 @@
...
+    />
+  );
+}
+
+const FrontLeadTwoAndTwo = (props) => {
+  return (
+    <ResponsiveSlice
+      renderSmall={(breakpoint, orientation) =>
+        renderMedium(props, breakpoint, orientation)
+      }
```
Geoff-Ford, line 84: Could you maybe add a function above the `return` here like:

```
const renderSlice = (breakpoint, orientation) => renderMedium(props, breakpoint, orientation);
```

Then each of these render props could just take the that function, like:

```
...
renderSmall={renderSlice}
...
```

Not a huge saving granted but saves some repetition as they are all using `renderMedium`

packages/edition-slices/src/slices/frontleadtwoandtwo/testSlice.js
```diff
@@ -0,0 +1,3703 @@
+export const testSlice = {
```
Geoff-Ford, line 1: Is this to go (see above comment)

packages/edition-slices/src/tiles/tile-al-front/index.js
```diff
@@ -0,0 +1,70 @@
...
+  onPress,
+  tile,
+  breakpoint = editionBreakpoints.wide,
+}) => {
+  const crop = getTileImage(tile, "crop32");
+  const styles = stylesFactory(breakpoint);
+
+  if (!crop) {
+    return null;
+  }
```
Geoff-Ford, line 27: One liner

packages/edition-slices/src/tiles/tile-d-front/index.js
```diff
@@ -0,0 +1,69 @@
...
+  onPress,
+  tile,
+  breakpoint = editionBreakpoints.small,
+}) => {
+  const crop = getTileImage(tile, "crop32");
+  const styles = stylesFactory(breakpoint);
+
+  if (!crop) {
+    return null;
+  }
```
Geoff-Ford, line 26: Guessing there might be a few of these 😝

packages/front-page/front-tile-summary.tsx
```diff
@@ -0,0 +1,114 @@
...
+  containerStyle?: any;
+  summary: Markup;
+  summaryStyle?: any;
+  tile: any;
+  bylines?: Markup;
+  whiteSpaceHeight: number;
+  linesOfTeaserToRender?: number | undefined;
+}
+
+class FrontTileSummary extends Component<Props> {
```
Geoff-Ford, line 27: 😱  Must this be done as a `class`?

packages/front-page/styles/index.js
```diff
@@ -0,0 +1,49 @@
...
+    ...textStyle,
+    fontSize: fontSizeLandscapeResolver[breakpoint],
+    lineHeight: lineHeightLandscapeResolver[breakpoint],
+  },
+  textPortrait: {
+    ...textStyle,
+    fontSize: fontSizePortraitResolver[breakpoint],
+    lineHeight: lineHeightPortraitResolver[breakpoint],
+  },
+  bylineContainer: { marginBottom: 5 },
```
Geoff-Ford, line 47: Should this be `spacing(1)`?

packages/slice-layout/src/templates/frontleadtwoandtwo/index.js
```diff
@@ -0,0 +1,111 @@
...
+  if (windowWidth >= 1112) {
+    return 20;
+  }
+  if (windowWidth >= 1080) {
+    return 15;
+  }
+
+  if (windowWidth >= 1024) {
+    return 15;
+  }
```
Geoff-Ford, line 46: All the above can be one liners no?

packages/front-page/front-article-summary-content/styles.js
```diff
@@ -0,0 +1,54 @@
...
+    ...textStyle,
+    fontSize: fontSizeResolver[breakpoint].landscape,
+    lineHeight: lineHeightResolver[breakpoint].landscape,
+  },
+  textPortrait: {
+    ...textStyle,
+    fontSize: fontSizeResolver[breakpoint].portrait,
+    lineHeight: lineHeightResolver[breakpoint].portrait,
+  },
+  bylineContainer: { marginBottom: 5 },
```
Geoff-Ford, line 53: Should use `spacing(1)`

packages/edition-slices/src/slices/frontleadtwoandtwo/index.js
c-j-j: Yes, exactly. This is the only way I could figure out how render the new slice in the app. The tiles don't render well on storybook as they get overfilled with text (hoping to look into that today)

packages/front-page/front-tile-summary.tsx
c-j-j: I copied/pasted the original TileSummary - I can change to a functional one

packages/front-page/styles/index.js
c-j-j: Yes good spot

packages/edition-slices/src/slices/frontleadtwoandtwo/testSlice.js
c-j-j: Yes

packages/front-page/front-article-summary-content/styles.js
aronshamash: yes sounds better

packages/front-page/front-tile-summary.tsx
c-j-j: Done

packages/edition-slices/src/tiles/tile-e-front/index.js
```diff
@@ -0,0 +1,79 @@
...
+  showSummary,
+  showByline,
+  breakpoint = editionBreakpoints.small,
+}) => {
+  const crop = getTileImage(tile, "crop45");
+  const styles = stylesFactory(breakpoint);
+
+  if (!crop) {
+    return null;
+  }
```
Geoff-Ford, line 29: Thought you might have found the other ones ;-)

PR #75, opened 2020-08-04
feat: Enable scrollbars on article page tablet only

aronshamash:
- Enable scrollbars on tablet devices on article page.

### DIFF
--- a/packages/article-skeleton/src/article-skeleton.js
+++ b/packages/article-skeleton/src/article-skeleton.js
@@ -144,7 +144,7 @@ const ArticleWithContent = (props) => {
           ListHeaderComponent={header}
           ListFooterComponent={Loading}
           onEndReached={onEndReached}
-          showsVerticalScrollIndicator={false}
+          showsVerticalScrollIndicator={isTablet}
           renderItem={Child}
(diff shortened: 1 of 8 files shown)


packages/article-skeleton/src/article-skeleton.js
```diff
@@ -144,7 +144,7 @@ const ArticleWithContent = (props) => {
           ListHeaderComponent={header}
           ListFooterComponent={Loading}
           onEndReached={onEndReached}
-          showsVerticalScrollIndicator={false}
+          showsVerticalScrollIndicator={isTablet}
```
manosim, line 147: Should we also bring them back for mobile as well? Do we know if there was a specific reason we disabled them in the first place?

aronshamash: Apparently PO would like to see what it looks like but just for tablet. I am going to look at why it was disabled in first place.

aronshamash: It was introduced in this monster PR: https://github.com/newsuk/times-components/pull/2330 to fix cut off article paragraphs

PR #77, opened 2020-08-06
chore: Setup Navigation

manosim:
### To Do
Use the React Native project as an "app" and not just for storybook. Prepare the project for E2E tests.

### Tasks

- [x] Move JS from `storybook-native/` to `app/`
- [x] Use TS for `app/`
- [x] Setup react-native-navigation
- [x] Setup tabs
- [x] Allow storybook to be run within the app
- [x] Rename scripts
- [x] Add a checkbox to the list and mark it as done


app/navigation/index.js
```diff
@@ -0,0 +1,19 @@
+import { Navigation } from "react-native-navigation";
+
+import * as ViewsIDs from "./ids";
+import { EditionView } from "../views/Edition";
+import { PastSixDaysView } from "../views/PastSixDays";
+import { MyArticlesView } from "../views/MyArticles";
+import { MoreView } from "../views/More";
+import { StorybookUIRoot } from "../storybook/";
+
+// prettier-ignore
```
manosim, line 10: This one was done on purpose. If one of the component names is a bit longer, the registered screens list won't be very readable.

android/build.gradle
```diff
@@ -3,10 +3,11 @@
 buildscript {
     ext {
         buildToolsVersion = "28.0.3"
-        minSdkVersion = 16
+        minSdkVersion = 19
         compileSdkVersion = 28
         targetSdkVersion = 28
         kotlin_version = '1.3.30'
```
Geoff-Ford, line 9: Maybe worth bumping this also? They are on `1.3.72` currently so nothing major but a lot of bug fixes.

app/navigation/index.js
```diff
@@ -0,0 +1,19 @@
+import { Navigation } from "react-native-navigation";
+
+import * as ViewsIDs from "./ids";
```
Geoff-Ford, line 3: Is there a need to export each ID separately above if you are going to just pull them all in together?

app/navigation/options.tsx
```diff
@@ -0,0 +1,37 @@
+import { Options, OptionsBottomTab } from "react-native-navigation";
+
+import { TIMES_BLUE } from "../utils/colors";
+import { GillSansMTStdMedium, TimesModern } from "../utils/fonts";
+
+export const getOptions = (title: string, opts: Options = {}): Options => {
+  return {
```
Geoff-Ford, line 7: Can implicit return here ` => ({`

app/navigation/options.tsx
```diff
@@ -0,0 +1,37 @@
...
+    ...opts,
+    topBar: {
+      ...opts.topBar,
+      background: {
+        color: "black",
+        ...(opts.topBar || {}).background,
```
Geoff-Ford, line 13: Can you use optional chaining here? `...opts.topBar?.background`

app/navigation/options.tsx
```diff
@@ -0,0 +1,37 @@
...
+    statusBar: {
+      style: "light",
+      ...opts.statusBar,
+    },
+  };
+};
+
+export const bottomTabOptions: OptionsBottomTab = {
+  iconColor: "#C2C2C2",
+  textColor: "#C2C2C2",
```
Geoff-Ford, line 32: Can `#C2C2C2` go into the `utils/colors` also?

app/views/Edition.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from "react";
```
Geoff-Ford, line 1: This can just be `import React from "react";`. Same on the other files. Or (my preference) just grab what you need `import React, { FC } from "react"

app/navigation/index.js
manosim: This one was done on purpose because you need the id of the destination for a specific view to be able to push to the nav stack. So from "home" to push to the "article", you need the VIEW_ARTICLE_ID (example in this PR in `app/views/Edition.tsx`). But in some files (like this one) you use multiple ones so instead of importing one by one, I pull them all in together.

app/navigation/options.tsx
manosim: This particular is not part of our brand, hence I kept it hardcoded as planning to bring our brand colors later on.

app/views/Edition.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { Navigation, NavigationComponentProps } from "react-native-navigation";
+import { Button, StyleSheet, View } from "react-native";
+
+import { VIEW_ID_STORYBOOK } from "../navigation/ids";
+
+const styles = StyleSheet.create({
```
Geoff-Ford, line 7: Minor thing but can you move this to the bottom. Just more of a convention than anything else.

app/views/More.tsx
```diff
@@ -0,0 +1,29 @@
...
+    alignItems: "center",
+    padding: 20,
+  },
+  title: {
+    fontFamily: GillSansMTStdMedium,
+    fontSize: 20,
+  },
+});
+
+export const MoreView: React.FC<NavigationComponentProps> = () => {
```
Geoff-Ford, line 21: These can be explicit return also.

app/views/Edition.tsx
```diff
@@ -0,0 +1,41 @@
...
+  container: {
+    flex: 1,
+    backgroundColor: "white",
+    justifyContent: "center",
+    alignItems: "center",
+    padding: 20,
+  },
+});
+
+export const EditionView: React.FC<NavigationComponentProps> = (props) => {
```
Geoff-Ford, line 17: Maybe one day, over a beer, we can discuss why you hate default exports so much 🤣

c-j-j: I'm convinced after working on our classic app. Makes it much easier to keep names consistent across a codebase. TimesComponents has a lot of examples where the imported name is different from the exported name as a default export was used.

Geoff-Ford: True enough. And you can still do an `as` if you need to call it something else, but you can still find it by original name in a search.

PR #80, opened 2020-08-07
chore(NO-JIRA): Send release notes to Slack

manosim:
image


scripts/send-slack-message.js
```diff
@@ -0,0 +1,90 @@
...
+  } catch (_) {
+    console.error("Something went wrong.");
+  }
+};
+
+const sendSlackMessage = async (notes, htmlUrl) => {
+  const parsedNotes = marked(notes.trim());
+
+  const body = {
+    text: "A new *Times Components* release is now _available_.",
```
c-j-j, line 40: Should this be `Times Components Native` to help differentiate between TCN and TC?

manosim: Good point!

PR #86, opened 2020-08-10
fix(TNLT-3679): Renders FrontSection slice without FlatList.

c-j-j:
We don't need to render a FlatList for the front page as it's a full-screen slice with no scrolling enabled.

Also reduces paragraph indentation

### DIFF
--- a/packages/front-page/front-renderer.js
+++ b/packages/front-page/front-renderer.js
@@ -2,7 +2,7 @@ import React from "react";
 import coreRenderers from "@times-components-native/markup";
 import { Text } from "react-native";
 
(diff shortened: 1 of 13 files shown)


packages/front-page/front-renderer.js
```diff
@@ -2,7 +2,7 @@ import React from "react";
 import coreRenderers from "@times-components-native/markup";
 import { Text } from "react-native";
 
-export const PARAGRAPH_INDENT = `\u3000\u3000`; // approximates a tab
+export const PARAGRAPH_INDENT = `\u3000`; // approximates a tab
```
aronshamash, line 5: 👍

PR #94, opened 2020-08-13
chore: add new queries for top secondary slices

aronshamash:
- TPA now provides support for  `TopSecondaryFourSlice`, `TopSecondaryTwoAndTwoSlice` and `TopSecondaryTwoNoPicAndTwoSlice`. 

This adds these slices to the query.


c-j-j:
🚫  Don't merge this as is. The query at the moment is missing fields for some of the secondary-top queries (i.e. we need to always query for 16:9 and 3:2 crops, and summary125 and summary800). We need to review again if TPA should only expose a single secondary-top slice that we can query for

Geoff-Ford (approved):

PR #96, opened 2020-08-14
feat: columns

c-j-j:
This PR mainly copies over the column logic from nuk-tnl-app-ui to the packages/article-columns directory, and changes the new front tile logic to include the ability to format using columns.

We were able to simplify some of the column logic from classic, as we don't need to support inline images/multiple pages/etc.

- [x] Render bylines inside column if columnCount > 1
- [x] ensure columns rerender correctly on orientation change (this is fixed by suffixing the height to the content ID)
- [x] figure out what we want to do with types (our new types are more defined than existing ones like Markup) - should these go in a more central place. **For now - we only use the ParagraphContent types in the column logic - and replacing these with the Markup types doesn't make much sense as we'll lose a lot of type information**

image


packages/front-page/front-page-byline/styles.ts
```diff
@@ -0,0 +1,11 @@
+import { spacing, colours } from "@times-components-native/styleguide";
+
+export default () => ({
+  bylineContainer: { marginBottom: spacing(1) },
+  bylineStyle: { lineHeight: 13 },
+  withKeyline: {
+    borderTopWidth: 1,
+    borderColor: colours.functional.darkGrey,
+    paddingTop: 5,
```
Geoff-Ford, line 9: `spacing(1)`?

packages/article-columns/types.ts
```diff
@@ -0,0 +1,39 @@
+import { ParagraphContent } from "@times-components-native/article-columns/domain-types";
```
Geoff-Ford, line 1: Missed one ;-)

PR #97, opened 2020-08-14
feat: Top secondary four slice

aronshamash:
- https://nidigitalsolutions.jira.com/browse/TNLT-3530
- Creation of new slice `TopSecondarySlice` which will be used when TPA return any of the following slice types:

- `TopSecondaryFourSlice`
- `TopSecondaryTwoNoPicAndTwoSlice`
- `TopSecondaryTwoAndTwoSlice`

This requires a transformation of slice items which should arguably happen in TPA, and it  could return just `TopSecondarySlice` for all three slices to avoid the transformation within the slice.

- New tiles `TileAT`, `TileAU` and `TileV` created.

### Portrait
Screenshot 2020-08-14 at 13 40 40

### Landscape
Screenshot 2020-08-14 at 13 42 15

In keeping with existing section slices, there is not styling based on orientation, it is purely done on breakpoint.


packages/edition-slices/__tests__/shared-tablet-slices.base.js
```diff
@@ -129,6 +130,21 @@ const slices = [
     name: "front lead two no pic and two",
     Slice: LeadTwoNoPicAndTwoFrontSlice,
   },
+  {
+    mock: mockSecondaryFourSlice(),
```
c-j-j, line 134: As discussed, we'll need to create new mocks for the new top slices here if we're keeping existing approach.

c-j-j: No need to do this anymore as we request a generic array of `items`, and we don't transform the slice based on the slice name

c-j-j:
Most notable changes that I've made:
1. The section_fragment.graphql query now just fetches an array of `items`. We no longer need to specify which tile has which fields.
2. The top-secondary slice now uses the orientation to decide which tiles to show instead of breakpoint. This is because we were previously showing a landscape design on iPad Pro 12.9'' portrait - and this generally simplifies things
3. Added landscape/portrait snapshots for secondary top, as well as new front slices.
4. Landscape lead tile (TileAU) now renders 16:9 instead of 3:2.

c-j-j:
Updated screenshots:

Simulator Screen Shot - iPad Pro (11-inch) (2nd generation) - 2020-08-20 at 16 38 08
Simulator Screen Shot - iPad Pro (11-inch) (2nd generation) - 2020-08-20 at 16 38 05

PR #104, opened 2020-08-20
chore: Added dynamic setting of dianomi context id to sponsored ads

Geoff-Ford:
… and added a keyline above.

Screenshot 2020-08-20 at 15 20 01

Screenshot 2020-08-20 at 15 49 35


packages/ad/src/sponsored-ad.tsx
```diff
@@ -12,17 +17,34 @@ const handleRequest = (e: WebViewNavigation) => {
...
+interface Props {
+  narrowContent?: boolean;
+}
+
+export const SponsoredAd: React.FC<Props> = ({ narrowContent = false }) => {
+  const contextId = narrowContent ? "244" : "243";
+  return (
+    <View
+      style={{
+        borderTopWidth: 1,
```
c-j-j, line 29: Whoops - looks like I didn't extract the styles here into a styles.ts file when I coded this. Do you mind doing that?

c-j-j (approved):

packages/ad/src/sponsored-ad.tsx
Geoff-Ford: Can do

Geoff-Ford: Done

PR #112, opened 2020-08-26
fix: fixes front section on android to not flow under navbar.

c-j-j:
Adds extra padding at the bottom of a front on android to factor in bottom nav height.

BEFORE:
image

AFTER:
image


Geoff-Ford (approved):

c-j-j:
This fix makes things almost worse on an nexus 9 as there's not enough room. Maybe best to close this PR and try to figure out a correct fix for this?
image

PR #118, opened 2020-08-28
feat: Supplement section logic

aronshamash:
- provide a temporary way to identify is a slice is within a supplement section.

### DIFF
--- a/packages/edition-slices/src/edition-slices.js
+++ b/packages/edition-slices/src/edition-slices.js
@@ -31,7 +31,7 @@ const isTablet =
   (config && config.breakpoint && config.breakpoint !== "small") ||
   width > tabletWidth;
 
-export const sliceMap = (isInSupplement) => {
+const sliceMap = (isInSupplement) => {
   const isInTabletSupplement = isInSupplement && isTablet;
   return {
     CommentLeadAndCartoonSlice,
--- a/packages/edition-slices/src/edition-slices.js
+++ b/packages/edition-slices/src/edition-slices.js
@@ -31,9 +31,9 @@ const isTablet =
   (config && config.breakpoint && config.breakpoint !== "small") ||
   width > tabletWidth;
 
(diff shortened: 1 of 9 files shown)


c-j-j (approved):

packages/edition-slices/src/edition-slices.js
```diff
@@ -30,33 +30,39 @@ const { width } = getDimensions();
...
-  LeadTwoNoPicAndTwoFrontSlice,
-  LeadOneAndOneFrontSlice,
-  LeadOneFullWidthFrontSlice,
-  TopSecondaryTwoAndTwoSlice: TopSecondarySlice,
-  TopSecondaryTwoNoPicAndTwoSlice: TopSecondarySlice,
-  TopSecondaryFourSlice: TopSecondarySlice,
+const getSlice = (isInSupplement, sliceName) => {
+  const SecondaryTwoAndTwoMapper = isTablet
+    ? SecondaryTwoNoPicAndTwoSlice
+    : SecondaryTwoAndTwoSlice;
```
Geoff-Ford, line 37: We should use either a separate mapper for each "different" like this one or inline them like you have on your one below but not both.

packages/edition-slices/src/edition-slices.js
```diff
@@ -30,33 +30,39 @@ const { width } = getDimensions();
...
+
+  const sliceMap = {
+    CommentLeadAndCartoonSlice,
+    DailyUniversalRegister: DailyRegisterLeadFourSlice,
+    LeadersSlice,
+    LeadOneAndFourSlice:
+      isInSupplement && isTablet ? LeadOneAndFourSlice : LeadOneAndFourSlice,
```
Geoff-Ford, line 44: Maybe put `isInSupplement && isTablet` together in a `const` like `isInTabletSupplement` as it will be reused by the other supplement templates

Geoff-Ford (approved):

PR #119, opened 2020-08-28
fix: supplement titles

aronshamash:
- Fix supplement titles

### DIFF
--- a/packages/section/src/utils.js
+++ b/packages/section/src/utils.js
@@ -108,10 +108,10 @@ const createPuzzleData = (puzzles, editionBreakpoint) => {
 
 const isSupplementSection = (sectionTitle) => {
   const knownSupplementSections = [
-    "Bricksmortar",
+    "Bricks & Mortar",
     "Style",
-    "Thesundaytimesmagazine",
-    "Thetimesmagazine",
+    "The Sunday Times Magazine",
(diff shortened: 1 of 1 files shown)


packages/section/src/utils.js
```diff
@@ -108,10 +108,10 @@ const createPuzzleData = (puzzles, editionBreakpoint) => {
 
 const isSupplementSection = (sectionTitle) => {
   const knownSupplementSections = [
-    "Bricksmortar",
+    "Bricks & Mortar",
```
c-j-j, line 111: I thought we said in refinement that we'll use section slugs for temporarily checking a supplement instead of section names - as they'll be less likely to change?

aronshamash: we could do that but it would require a query change too. I thought as this is only a temporary measure it would be ok to check the title, as we already have that. Checking the slug will not capture any more supplement sections than checking title, only advantage would be if say `Bricks & Mortar` changed to `Bricks and Mortar`, realistically not likely to happen. What do you think ?

c-j-j: if this is temporary, this is probably okay for beta... 

as you say - hopefully, it'll really unlikely they'll change one of these section names

c-j-j (approved):

PR #124, opened 2020-09-04
feat: Supplement secondary two and two

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-4215

**iPad Pro (12.9inch) Portrait**
Screenshot 2020-09-04 at 14 25 42

**iPad Pro (12.9inch) Landscape**
Screenshot 2020-09-04 at 14 39 39


packages/provider-queries/src/section_fragment.graphql
```diff
@@ -682,33 +682,33 @@ fragment commonSlicesSection on StandardSectionSlice {
     secondary1 {
```
c-j-j, line 682: TileBE needs a 1:1 crop - do we need to add that here, or is it already queried for?

c-j-j (approved):

packages/provider-queries/src/section_fragment.graphql
Geoff-Ford: Tiles BD and BE are only for the support articles which are already covered in the query:
https://github.com/newsuk/times-components-native/blob/2c975e00ca81c99f1cdaad7b747ea390db1d9707/packages/provider-queries/src/section_fragment.graphql#L719

aronshamash (approved):

PR #128, opened 2020-09-08
chore: update article page query to include topics

aronshamash:
-  update article page query to include topics
- use separate files for mobile and tablet to allow native to use approach query based on device type.

- [x] Test ios works as expected
- [ ] Test Android works as expected


Geoff-Ford (approved):

c-j-j (approved):

aronshamash:
@Geoff-Ford @c-j-j can we leave this as reference but we when we get to play https://nidigitalsolutions.jira.com/browse/TNLT-4868

PR #130, opened 2020-09-10
fix: (TNLT-4746) Add support for lead asset in tablet comment template

aronshamash:
- Added support for lead asset in the comment tablet template but only for articles with the `magazinecomment` templateid to avoid duplicate author images being displayed.

### iPad
Screenshot 2020-09-10 at 15 34 33

Screenshot 2020-09-10 at 15 53 10

### iPad Pro
Screenshot 2020-09-10 at 16 14 44

Screenshot 2020-09-10 at 16 15 09

### DIFF
--- a/packages/article-comment-tablet/__tests__/shared-tablet.native.js
+++ b/packages/article-comment-tablet/__tests__/shared-tablet.native.js
@@ -29,13 +29,34 @@ export default () => {
 
   const tests = [
     {
-      name: "Article Comment Tablet",
+      name: "Article Comment Tablet - magazinecomment template",
       test() {
         const testInstance = TestRenderer.create(
           withTabletContext(
             <ArticleCommentTablet
               {...sharedProps}
-              article={articleFixture()}
+              article={{
+                template: "magazinecomment",
+                ...articleFixture(),
+              }}
+            />,
+          ),
+        );
+
+        expect(testInstance).toMatchSnapshot();
+      },
+    },
+    {
+      name: "Article Comment Tablet - maincomment template",
+      test() {
+        const testInstance = TestRenderer.create(
+          withTabletContext(
+            <ArticleCommentTablet
+              {...sharedProps}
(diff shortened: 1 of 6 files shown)


packages/article-comment-tablet/src/article-comment-tablet.js
```diff
@@ -33,21 +39,37 @@ class ArticlePage extends Component {
...
-        headline={getHeadline(headline, shortHeadline)}
-        label={label}
-        longRead={longRead}
-        onAuthorPress={onAuthorPress}
-        onImagePress={onImagePress}
-        publicationName={publicationName}
-        publishedTime={publishedTime}
-        standfirst={standfirst}
-      />
+      <Fragment>
```
Geoff-Ford, line 47: Can you use `<></>` here?

packages/article-comment-tablet/__tests__/shared-tablet.native.js
```diff
@@ -29,13 +29,34 @@ export default () => {
...
+      name: "Article Comment Tablet - magazinecomment template",
       test() {
         const testInstance = TestRenderer.create(
           withTabletContext(
             <ArticleCommentTablet
               {...sharedProps}
-              article={articleFixture()}
+              article={{
+                template: "magazinecomment",
+                ...articleFixture(),
```
Geoff-Ford, line 40: I can see this is working as the snapshots are correctly showing the `ArticleLeadAsset` and I see the `full-article` fixture doesn't have a `template` key. But might be worth putting the `template` after the spread to ensure it will always override the fixture incase someone adds a `template` at some point. Same for the other test below.

Geoff-Ford (approved):

packages/article-comment-tablet/src/article-comment-tablet.js
aronshamash: will do

PR #131, opened 2020-09-11
fix: (TNLT-4760) Inline image layout with narrow content

aronshamash:
- Fixes Inline image layout with narrow content templates

### Before
Screenshot 2020-09-11 at 15 54 27

### After
Screenshot 2020-09-11 at 15 48 04

### DIFF
--- a/packages/article-image/src/article-image.js
+++ b/packages/article-image/src/article-image.js
@@ -21,7 +21,10 @@ const ArticleImageNative = (props) => {
             narrowContent && [
               styles[`${display}ContainerNarrow`],
               {
-                width: narrowArticleBreakpoint.content,
+                width:
+                  display !== "inline"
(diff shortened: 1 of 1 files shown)


packages/article-image/src/article-image.js
```diff
@@ -21,7 +21,10 @@ const ArticleImageNative = (props) => {
             narrowContent && [
               styles[`${display}ContainerNarrow`],
               {
-                width: narrowArticleBreakpoint.content,
+                width:
+                  display !== "inline"
+                    ? narrowArticleBreakpoint.content
+                    : "auto",
```
Geoff-Ford, line 27: Can these width alternatives not be expressed in their respective styles i.e. `inlineContainerNarrow`, `primaryContainerNarrow`, `secondaryContainerNarrow`

c-j-j (approved):

PR #132, opened 2020-09-11
fix: Removes inline image if it is the first item in the article content

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-4684

**Before:**
Screenshot 2020-09-11 at 18 48 19

**After:**
Screenshot 2020-09-11 at 18 49 42

PS: Added `Dimensions` to our `modulePaths` as it was missing.

### DIFF
--- a/lib/android/modulePaths.js
+++ b/lib/android/modulePaths.js
@@ -93,6 +93,7 @@ module.exports = [
   "node_modules/react-native/Libraries/Performance/Systrace.js",
   "node_modules/react-native/Libraries/StyleSheet/flattenStyle.js",
   "node_modules/react-native/Libraries/TurboModule/TurboModuleRegistry.js",
+  "node_modules/react-native/Libraries/Utilities/Dimensions.js",
   "node_modules/react-native/Libraries/Utilities/GlobalPerformanceLogger.js",
   "node_modules/react-native/Libraries/Utilities/HMRClient.js",
   "node_modules/react-native/Libraries/Utilities/NativePlatformConstantsAndroid.js",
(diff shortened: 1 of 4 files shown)


yarn.lock
```diff
@@ -85,7 +85,7 @@
     lodash "^4.17.13"
     source-map "^0.5.0"
 
-"@babel/generator@^7.11.0", "@babel/generator@^7.5.0":
+"@babel/generator@^7.11.0":
```
c-j-j, line 88: Just double checking - did you mean to push an updated yarn.lock?

c-j-j (approved):

lib/android/modulePaths.js
```diff
@@ -93,6 +93,7 @@ module.exports = [
   "node_modules/react-native/Libraries/Performance/Systrace.js",
   "node_modules/react-native/Libraries/StyleSheet/flattenStyle.js",
   "node_modules/react-native/Libraries/TurboModule/TurboModuleRegistry.js",
+  "node_modules/react-native/Libraries/Utilities/Dimensions.js",
```
aronshamash, line 96: I guess this shouldn't be here.

aronshamash (changes requested):

lib/android/modulePaths.js
Geoff-Ford: Yep, it should. It was missing before.

yarn.lock
Geoff-Ford: Updated yarn to 1.22.5, double checked on right node version (10.13.0) via nvm, deleted and regenerated `yarn.lock`

aronshamash (approved):

PR #134, opened 2020-09-14
fix(TNLT-4716): Fixes byline text wrapping

c-j-j:
https://nidigitalsolutions.jira.com/browse/TNLT-4716

We need to wrap byline in a Text component, which ensures that the nested Text components wrap to multiple lines as expected. 

**Before:**
image

**After:** 
image

### DIFF
--- a/packages/article-byline/src/render-byline.js
+++ b/packages/article-byline/src/render-byline.js
@@ -28,7 +28,11 @@ const renderByline = (Component, ast, textStyle, props = {}) => {
   const bylineAst = ast.map((bylineObj) =>
     bylineObj.byline && bylineObj.byline.length > 0 ? bylineObj.byline[0] : {},
   );
-  return renderTrees(bylineAst, bylineRenderers(Component, textStyle, props));
+  const trees = renderTrees(
+    bylineAst,
+    bylineRenderers(Component, textStyle, props),
+  );
+  return <Text>{trees}</Text>;
 };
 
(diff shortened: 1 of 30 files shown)


aronshamash (approved):

packages/article-byline/src/render-byline.js
```diff
@@ -28,7 +28,11 @@ const renderByline = (Component, ast, textStyle, props = {}) => {
   const bylineAst = ast.map((bylineObj) =>
     bylineObj.byline && bylineObj.byline.length > 0 ? bylineObj.byline[0] : {},
   );
-  return renderTrees(bylineAst, bylineRenderers(Component, textStyle, props));
+  const trees = renderTrees(
+    bylineAst,
+    bylineRenderers(Component, textStyle, props),
+  );
+  return <Text>{trees}</Text>;
```
Geoff-Ford, line 35: That's it??? haha

Geoff-Ford (approved):

packages/article-byline/src/render-byline.js
c-j-j: Yep... before we would have had something like `<View><Text>Hello</Text><Text>, World</Text>L</View>` - which will push the entire second Text component to a new line if there's not enough space. Changing this to `<View><Text><Text>Hello</Text><Text>, World</Text></Text></View>` seems to do the right thing.

PR #135, opened 2020-09-14
fix: (TNL-4681) Fix label padding in in-depth template on tablet

aronshamash:
- Fix label padding in in-depth template on tablet
- Adjust padding on header container to centrally align content
- https://nidigitalsolutions.jira.com/browse/TNLT-4681

### Before 
Screenshot 2020-09-14 at 12 31 39

### Before (no standfirst)
Screenshot 2020-09-14 at 12 31 53

### After 
Screenshot 2020-09-14 at 12 44 27

### After (no standfirst)
Screenshot 2020-09-14 at 12 44 50


Geoff-Ford:
Haven't had lunch yet. Those screenshots have me drooling!

c-j-j (approved):

packages/article-in-depth/src/article-label/article-label.js
```diff
@@ -9,13 +9,13 @@ import {
...
 
-const HeaderLabel = ({ color, isVideo, label }) => {
+const HeaderLabel = ({ color, isVideo, isTablet, label }) => {
   if (!isVideo && !label) return null;
 
   const Label = isVideo ? VideoLabel : ArticleLabel;
 
   return (
-    <View style={styles.label}>
+    <View style={[isTablet && styles.labelTablet]}>
```
Geoff-Ford, line 18: So the original `tablet` styles have been binned as only applying a `marginBottom` of 0 which is default?

Geoff-Ford (approved):

packages/article-in-depth/src/article-label/article-label.js
aronshamash: Yes is the answer! Basically I refactored the shared and index files, tidied up and applied fix for this bug. The label only needs margin for tablet not for mobile.

PR #136, opened 2020-09-14
feat(TNLT-4855): Hide save star on tablet

c-j-j:
https://nidigitalsolutions.jira.com/browse/TNLT-4855

### DIFF
--- a/packages/edition-slices/src/tiles/shared/positioned-tile-star.js
+++ b/packages/edition-slices/src/tiles/shared/positioned-tile-star.js
@@ -1,5 +1,5 @@
 /* eslint-disable react/require-default-props */
-import React from "react";
+import React, { useContext } from "react";
 import { View } from "react-native";
 import PropTypes from "prop-types";
 
@@ -11,6 +11,7 @@ import {
   starCenterStyles,
   starUnderneathTextStyles,
 } from "./styles";
+import { ResponsiveContext } from "@times-components-native/responsive";
 
 const PositionedTileStar = ({
   centeredStar = false,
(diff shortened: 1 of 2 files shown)


aronshamash:
@c-j-j no snap updates ?

c-j-j:
@aronshamash no - there are currently no unit tests for the PositionedTileStar itself, and it is mocked everywhere else. Happy to add unit tests for the star if it'll help - what do you think?

aronshamash:
> @aronshamash no - there are currently no unit tests for the PositionedTileStar itself, and it is mocked everywhere else. Happy to add unit tests for the star if it'll help - what do you think?

Lets leave as is for now.

aronshamash (approved):

packages/edition-slices/src/tiles/shared/positioned-tile-star.js
```diff
@@ -11,14 +11,17 @@ import {
...
+import { ResponsiveContext } from "@times-components-native/responsive";
 
 const PositionedTileStar = ({
   centeredStar = false,
   underneathTextStar = false,
   customStyles = {},
   ...props
 }) => {
-  if (!isSaveSupported) return null;
+  const { isTablet } = useContext(ResponsiveContext);
```
Geoff-Ford, line 22: Should we follow that pattern of creating a `useResponsiveContext` method to `@times-components-native/responsive`? What do you think?

Geoff-Ford (approved):

packages/edition-slices/src/tiles/shared/positioned-tile-star.js
c-j-j: Sounds good - I can do that in another PR - and then update all places that have `useContext(ResponsiveContext)`?

PR #154, opened 2020-09-21
fix(TNLT-4833): renders unknown AST correctly

c-j-j:
image

### DIFF
--- a/packages/article-skeleton/src/article-body/article-body-row.js
+++ b/packages/article-skeleton/src/article-body/article-body-row.js
@@ -373,8 +373,8 @@ export default ({
         </View>
       );
     },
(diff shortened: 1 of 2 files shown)


packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -373,8 +373,8 @@ export default ({
         </View>
       );
     },
-    unknown(key, attributes, children, index, tree) {
-      return tree;
+    unknown(key, attributes, children) {
+      return AttributedString.join(children);
```
aronshamash, line 377: yes this is same fix I had.

aronshamash (approved):

PR #157, opened 2020-09-23
chore(TNLT-4891): improve article skeleton performance

c-j-j:
A memory leak in the iOS app has now been fixed, which occurred when swiping across articles. So we've been able to remove some of the previous optimisations which slowed down the article page.

1. On the inline paragraph, we're now memoizing the result from the layout manager. `layoutManager.layout()` requires measuring each word in a paragraph - which is fairly slow.
2. When measuring font, we were also loading in the font at render time - which took around 20ms per font. I've changed this to be when the font is registered at startup time.
3. On the simple paragraph, we're memorizing the transformed text so it's ready for render. 
4. On the FlatList, we're setting removeClippedSubviews to false. This results in interactives etc not being unmounted when they're out of view. More info found here https://reactnative.dev/docs/optimizing-flatlist-configuration#removeclippedsubviews
5. On the FlatList, we've removed initialNumToRender (so it defaults back to 10). This results in seeing 10 paragraphs before they are lazily rendered, which is large enough for most articles.


Geoff-Ford (approved):

packages/article-skeleton/src/article-body/inline-paragraph.js
```diff
@@ -55,14 +51,23 @@ const InlineParagraph = ({
...
-  );
+    const newPositionedTextItems = manager.layout();
+    const newPositionItemSettings = newPositionedTextItems.map((p) =>
+      p.text.collapsedAttributes(0),
+    );
+    return [newPositionedTextItems, newPositionItemSettings];
+  }, [inlineExclusion]);
 
-  const positioned = manager.layout();
+  if (!str.length) {
```
aronshamash, line 68: why the later return ?

c-j-j: All hooks, including useMemo, need to be above conditional logic like this

PR #164, opened 2020-09-25
feat: (TNLT-5066) In Today's Edition component

aronshamash:
- In Today's Edition as a standalone component : `InTodaysEdition`
- Expanded API of `ForwardArrowIcon` to allow height to be passed

### iPad
Screenshot 2020-09-29 at 17 04 06

### iPad Pro 11
Screenshot 2020-09-29 at 17 16 21

### IPad Pro 12.9
Screenshot 2020-09-29 at 17 47 33

### iPad
ipad-proper-landscape

### iPad Air
Screenshot 2020-09-29 at 21 19 32

### iPad Pro 11
Screenshot 2020-09-29 at 21 20 19

### iPad Pro 12.9
Screenshot 2020-09-29 at 17 47 22


c-j-j (approved):
Looking good to me

packages/in-todays-edition/item.tsx
```diff
@@ -0,0 +1,63 @@
...
+  onArticlePress: <T = unknown, R = unknown>(args?: T) => R;
+  onLinkPress: <T = unknown, R = unknown>(args?: T) => R;
+  orientation: string;
+}
+
+const isArticleLink = (
+  link: ArticleLinkType | LinkType,
+): link is ArticleLinkType => {
+  // eslint-disable-next-line no-prototype-builtins
+  return link.hasOwnProperty("articleId");
```
Geoff-Ford, line 23: Could you do this with something like `return !!link.articleId;` Then you can avoid linting also and avoid the whole prototype thing

packages/in-todays-edition/item.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <Text style={styles.itemTitle}>{item.title}</Text>
+        <Text style={styles.itemStrapline}>{item.strapline}</Text>
+        <View style={styles.itemCTA}>
+          <Text style={styles.itemCTAText}>{ctaText}</Text>
+          <View style={styles.itemCTAIconContainer}>
+            <IconForwardArrow fillColour={colours.functional.red} height={8} />
+          </View>
+        </View>
+      </Link>
+      {index !== 3 ? <View style={styles.divider}></View> : null}
```
Geoff-Ford, line 58: What did we agree in the end? That the following is fine?
`{index !== 3 && <View style={styles.divider}></View>}`

packages/in-todays-edition/item.tsx
aronshamash: that what I had originally. If I do that we have this issue:
```
Property 'articleId' does not exist on type 'ArticleLinkType | LinkType'.
  Property 'articleId' does not exist on type 'LinkType'.ts(2339)
```

packages/in-todays-edition/item.tsx
aronshamash: ah yes maybe we did.

aronshamash: updated

packages/in-todays-edition/item.tsx
Geoff-Ford: As long as it doesn't have a knock on effect, think I'd rather have a standard approach and tslint comment than coding around TS problem but still needing a comment anyway.

packages/in-todays-edition/item.tsx
Geoff-Ford: Looks like there are two updates. Once changing the above and another changing it back

packages/in-todays-edition/item.tsx
aronshamash: ok agreed. updated now

packages/in-todays-edition/in-todays-edition.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+interface Props {
+  items: [ItemType];
+  onArticlePress: <T = unknown, R = unknown>(args?: T) => R;
+  onLinkPress: <T = unknown, R = unknown>(args?: T) => R;
+  orientation: string;
+}
+
+const text = {
+  heading: "IN TODAY'S EDITION",
```
Geoff-Ford, line 33: Sorry, just spotted this. I guess you thought there might be more static text? As there is only one thing, can this not just be a const like `headingText` rather than an object?

aronshamash: was going to put other text in there but broke out into separate file now.

Geoff-Ford (approved):

PR #170, opened 2020-09-28
chore: Adds device style util

c-j-j:

### DIFF
--- a/packages/styleguide/src/breakpoints/index.js
+++ b/packages/styleguide/src/breakpoints/index.js
@@ -40,13 +40,10 @@ const getNarrowArticleBreakpoint = (width) => {
 };
 
 const getStyleByDeviceSize = (styles, windowWidth) => {
-  const sizes = Object.keys(styles)
-    .map(Number)
-    .sort((a, b) => b - a);
-
-  const index = sizes.findIndex((size) => size <= windowWidth);
-  const selectedSize = sizes[index];
-  return styles[`${selectedSize}`];
+  const selectedSize = Object.entries(styles)
+    .sort(([a], [b]) => b - a)
+    .find(([size]) => size <= windowWidth);
+  return selectedSize?.[1] ?? {};
 };
 
 export {
--- a/packages/styleguide/src/breakpoints/index.js
+++ b/packages/styleguide/src/breakpoints/index.js
@@ -42,8 +42,7 @@ const getNarrowArticleBreakpoint = (width) => {
 const getStyleByDeviceSize = (styles, windowWidth) => {
   const sizes = Object.keys(styles)
     .map(Number)
-    .sort((a, b) => a > b)
-    .reverse();
+    .sort((a, b) => b - a);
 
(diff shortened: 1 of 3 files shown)


packages/styleguide/src/breakpoints/index.js
```diff
@@ -39,13 +39,25 @@ const getNarrowArticleBreakpoint = (width) => {
   return narrowArticleWidths.wide;
 };
 
+const getStyleByDeviceSize = (styles, windowWidth) => {
+  const sizes = Object.keys(styles)
+    .map(Number)
```
Geoff-Ford, line 44: Is there a reason you don't just key the style groupings with a number rather than a string?

packages/styleguide/src/breakpoints/index.js
```diff
@@ -39,13 +39,25 @@ const getNarrowArticleBreakpoint = (width) => {
...
+    .sort((a, b) => a > b)
```
Geoff-Ford, line 45: I like this as you can't trust humans

aronshamash (approved):

packages/styleguide/src/breakpoints/index.js
Geoff-Ford: Actually shouldn't this be:
`.sort((a, b) => a - b)`
Not sure it will be sorted correctly as it is. Also, using `-` it should sort the array the right way around without needing to `reverse`?

packages/styleguide/src/breakpoints/index.js
```diff
@@ -39,13 +39,25 @@ const getNarrowArticleBreakpoint = (width) => {
...
+    .reverse();
+
+  const index = sizes.findIndex((size) => size <= windowWidth);
+  const selectedSize = sizes[index];
+  return styles[`${selectedSize}`];
+};
+
```
Geoff-Ford, line 51: I did something similar in my own project. How about something like this?

```
  const selectedSize = Object.entries(styles).sort(([a], [b]) => a - b).find(
    ([size]) => size <= windowWidth
  );
  return selectedSize?.[0] ?? {};
```

c-j-j: I can change this so it's one line - but not a massive fan of how you need to destructure the result of Object.entries. `sort(([a], [b]))` and `find([size])` looks a bit odd to me.

packages/styleguide/src/breakpoints/index.js
Geoff-Ford: We can still leave as strings as they will be coerced to numbers when compared with another number

packages/styleguide/src/breakpoints/index.js
Geoff-Ford: Also returning a default empty object so trying access `foo.myClassname` won't throw if no match found, it'll just return `undefined`.

c-j-j: Sounds good to me - prevents adding null checking everywhere.

PR #171, opened 2020-09-30
fix: decreases initial number of paragraphs 

c-j-j:
Inline paragraphs are very slow to render. As we're initially rendering 10 paragraphs - it's likely that lots of these will be inline paragraphs - which causes a janky experience when horizontally scrolling.


c-j-j:
@aronshamash @Geoff-Ford maybe we just go back to 2 paragraphs to prevent inline paragraphs from being rendered whilst scrolling horizontally through articles?

Geoff-Ford:
> @aronshamash @Geoff-Ford maybe we just go back to 2 paragraphs to prevent inline paragraphs from being rendered whilst scrolling horizontally through articles?

Which gets us back to the original trade off. Either we have performant swiping but a visible load of inline items like the ads, or we have the inline ads pre-rendered by including enough paragraphs to reach them (5?) for a smooth article scroll down, but at the cost of the article swipe. Rock and hard place.

c-j-j:
@Geoff-Ford changed the paragraph count to 4 - which will prevent an inline paragraph from being rendered

Geoff-Ford (approved):

aronshamash (approved):

PR #182, opened 2020-10-09
chore: Some performance improvements to try out

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-5539

Some performance improvements to try out for Ads, FlatList, and inlineParagraphs

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "1.10.7-beta",
+  "version": "1.10.8-beta",
   "description": "A collection of react-native components for The Times and Sunday Times",
   "main": "index.js",
   "engines": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "1.10.6",
+  "version": "1.10.7-beta",
   "description": "A collection of react-native components for The Times and Sunday Times",
   "main": "index.js",
   "engines": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "1.10.8-beta",
+  "version": "1.10.9-beta",
   "description": "A collection of react-native components for The Times and Sunday Times",
   "main": "index.js",
   "engines": {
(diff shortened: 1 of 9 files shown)


aronshamash (approved):
did you check that android is still working roughly same as before ?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "1.10.6",
+  "version": "1.10.7-beta",
```
c-j-j, line 4: reminder to revert this when you're ready to merge

packages/article-skeleton/src/body-utils.js
```diff
@@ -24,7 +24,8 @@ export const collapsed = (isTablet, content) =>
           // forwards
           let i;
           let children = [node];
-          for (i = 0; i < acc.length; i += 1) {
+          const maxNodesToInline = Math.min(4, acc.length);
```
c-j-j, line 27: Just to clarify, this is going to render a max of 4 paragraphs next to an inline image/ad? Sounds like it'll be enough for most of the time - but do you know what'll happen if it isn't enough?

philipe28: You get white space. CW and I looked across multiple examples and decided that 4 paras should cover most cases for double mpu ads and he was happy with a little white space if it happens.

Geoff-Ford: Yep, you will see some white space for the remainder of the height of the inline element. I ran through multiple articles with CW to check and he is happy with a little white space if 4 paras doesn't cover the height. Most likely this will only happen on Double MPUs.

c-j-j: And it seems that Double MPUs are fairly unlikely to be used - so shouldn't be too much of an issue.

packages/ad/src/dom-context.js
```diff
@@ -43,6 +44,10 @@ class DOMContext extends PureComponent {
       readableVersion: DeviceInfo.getReadableVersion(),
       version: DeviceInfo.getVersion(),
     };
+
+    setTimeout(() => {
+      this.setState({ ready: true });
+    }, 500);
```
c-j-j, line 50: Just out of interest, what's the reason for delaying an ad?

PR #183, opened 2020-10-12
chore: (TNLT-4599) Add new font for front bylines

aronshamash:
- https://nidigitalsolutions.jira.com/browse/TNLT-4599
- New font `gillsansmtstd-bold` added for use with front bylines. 
- Supplied ios and android app with font to add to respective repos.
- Also need a render pr to host font on public font folder on www.thetimes.co.uk for storybook 
see https://github.com/newsuk/cps-content-render/pull/4273


c-j-j (approved):
Do we need to add the new font to both  the `develop` and `UA` branches?

aronshamash:
> Do we need to add the new font to both the `develop` and `UA` branches?

should be only `UA` branches because only used on tablet fronts.

PR #184, opened 2020-10-13
feat(TNLT-5525): Dynamic front tiles 

c-j-j:
When rendering the front page summary, we now do the following:
1. Pre-render the headline/strapline/byline into the tile container to check how high each will be.
2. Measure the height of the tile container
3. Work out which text strings can fit into the container based on their priority - this information is worked out by the `getFrontTileConfig` helper
4. Render text strings.  The last text string in the tile summary will have its marginBottom set to 0, which is to ensure that text can use up all available space.

**Order of priority**
Headline > Strapline > Byline > Content

**With strapline**
Simulator Screen Shot - iPad Air (3rd generation) - 2020-10-16 at 11 16 52
Simulator Screen Shot - iPad Air (3rd generation) - 2020-10-16 at 11 17 04

**Without strapline**
Simulator Screen Shot - iPad Air (3rd generation) - 2020-10-16 at 11 14 20
Simulator Screen Shot - iPad Air (3rd generation) - 2020-10-16 at 11 15 18
Simulator Screen Shot - iPad Air (3rd generation) - 2020-10-16 at 11 15 34

### DIFF
--- a/packages/edition-slices/edition-front.showcase.js
+++ b/packages/edition-slices/edition-front.showcase.js
@@ -12,11 +12,20 @@ import {
   LeadTwoFrontSlice,
   LeadOneFullWidthFrontSlice,
 } from "./src/slices";
+import { getDimensions } from "@times-components-native/utils";
 
+const topNavHeight = 178;
+const bottomNavHeight = 48;
 const renderSlice = (Component, data) => () => {
+  const { height } = getDimensions("window");
   return (
     <Responsive>
-      <View style={{ flex: 1 }}>
+      <View
+        style={{
+          width: "100%",
+          height: height - topNavHeight - bottomNavHeight,
+        }}
+      >
         {
           <Component
             onPress={() => null}
--- a/packages/edition-slices/src/tiles/tile-f-front/styles/index.js
+++ b/packages/edition-slices/src/tiles/tile-f-front/styles/index.js
@@ -19,6 +19,7 @@ const summary = {
 const strapline = {
   fontFamily: fonts.headlineRegular,
   color: colours.functional.primary,
+  marginBottom: spacing(3),
 };
 
 const sharedLandscapeStyles = {
--- a/packages/edition-slices/src/tiles/tile-f-front/styles/index.js
+++ b/packages/edition-slices/src/tiles/tile-f-front/styles/index.js
@@ -19,7 +19,6 @@ const summary = {
 const strapline = {
   fontFamily: fonts.headlineRegular,
   color: colours.functional.primary,
-  marginBottom: spacing(3),
 };
 
 const sharedLandscapeStyles = {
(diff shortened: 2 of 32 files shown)


packages/edition-slices/edition-front.showcase.js
```diff
@@ -12,11 +12,20 @@ import {
   LeadTwoFrontSlice,
   LeadOneFullWidthFrontSlice,
 } from "./src/slices";
+import { getDimensions } from "@times-components-native/utils";
 
+const topNavHeight = 178;
```
aronshamash, line 17: this is the same for both ios and android?

packages/edition-slices/src/tiles/tile-f-front/styles/index.js
```diff
@@ -173,13 +165,8 @@ const styles = {
...
         fontSize: 24,
         lineHeight: 24,
-        marginBottom: spacing(3),
-      },
-      summary: {
-        ...summary,
-        fontSize: 14,
-        lineHeight: 18,
       },
+      straplineMarginBottom: spacing(3),
```
aronshamash, line 169: what do you think about doing `straline.marginBottom` to be more consistent even-though only 1 styling rule ?

Geoff-Ford: Did you have to take a p?

aronshamash (approved):
looks good.

packages/edition-slices/src/tiles/tile-f-front/styles/index.js
aronshamash: maybe

c-j-j: The thing is that these aren't really `style` objects, they are more for the configuration that's used when calculating what can fit into a tile... 

My concern would be someone adding more styles to that `straline` object

packages/edition-slices/edition-front.showcase.js
c-j-j: unsure to be honest - I picked this out of the designs - which may not match the build exactly. Also - this is missing the android soft nav at the bottom, which may not be ideal.

I wanted the fronts in storybook to have roughly the same dimensions as what they'd be rendered in the app - which made it a lot easier to switch between different devices/different fronts.

packages/front-page/utils/get-front-tile-config.ts
```diff
@@ -0,0 +1,70 @@
...
+  const canAccommodateStrapline =
+    container.height >=
+    headline.height + headline.marginBottom + strapline.height;
+
+  const canAccommodateByline =
+    container.height >=
+    headline.height +
+      headline.marginBottom +
+      strapline.height +
+      strapline.marginBottom +
```
Geoff-Ford, line 36: So if there is no strapline will the above two values be passed in as `0`?

c-j-j: Hmmm good spot - height will be 0 but marginBottom will likely be set.

Geoff-Ford: Yep, so need to dynamically set it to either the value or 0 when passing in depending on if strapline exists

c-j-j: I did do that originally... but I changed it back as I wanted to avoid all tiles having to pass in `straplineMarginBottom={strapline ? straplineMarginBottom : 0}` - same will need to be done for the byline.

I can change `getFrontTileConfig`   to not factor in strapline/byline margins if they have 0 height (i.e. they are missing from the data).

c-j-j: @Geoff-Ford pushed a fix

Geoff-Ford (approved):

PR #189, opened 2020-10-15
feat: (TNLT-5500) Integrate fronts

aronshamash:
- Add `InTheNewsSlice` to front page query
- Fix some incorrect image crops in query
- on section page refactor `onArticlePress` to only send `id` as agreed with Native devs
- on section page refactor `onLinkPress` to only send `url` as agreed with Native devs

### Lead Two 
Screenshot 2020-10-15 at 10 44 02
Screenshot 2020-10-15 at 10 43 56

### Lead One and One
Screenshot 2020-10-15 at 15 38 57
Screenshot 2020-10-15 at 15 39 36

### Lead One
Screenshot 2020-10-15 at 17 01 51
Screenshot 2020-10-15 at 17 03 22

### Android
Screenshot 2020-10-16 at 15 18 36

### DIFF
--- a/packages/edition-slices/src/slices/frontleadone/index.js
+++ b/packages/edition-slices/src/slices/frontleadone/index.js
@@ -9,8 +9,12 @@ function renderMedium(props, breakpoint, orientation) {
     onPress,
     onLinkPress,
     slice: { lead },
-    inTodaysEditionSlice: { items: inTodaysEditionItems },
+    inTodaysEditionSlice,
   } = props;
+
+  if (!inTodaysEditionSlice?.items.length) return null;
+
+  const { items: inTodaysEditionItems } = inTodaysEditionSlice;
 
   return (
     <FrontLeadOneSlice
--- a/packages/edition-slices/src/slices/frontleadone/index.js
+++ b/packages/edition-slices/src/slices/frontleadone/index.js
@@ -9,12 +9,8 @@ function renderMedium(props, breakpoint, orientation) {
     onPress,
     onLinkPress,
     slice: { lead },
-    inTodaysEditionSlice,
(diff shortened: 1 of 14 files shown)


packages/edition-slices/src/slices/frontleadone/index.js
```diff
@@ -9,9 +9,13 @@ function renderMedium(props, breakpoint, orientation) {
     onPress,
     onLinkPress,
     slice: { lead },
-    inTodaysEditionSlice: { items: inTodaysEditionItems },
+    inTodaysEditionSlice,
   } = props;
 
+  if (!inTodaysEditionSlice?.items.length) return null;
```
c-j-j, line 15: What's the likelihood of inTodaysEdtionSlice coming back missing from TPA? 

I wonder if we should still render the Front slice but with a blank space where ITE component should be. I think that'll be more desirable than rendering no front. 

This might be okay as is for beta - but would be good to review how to handle this before public release.

c-j-j (approved):

packages/edition-slices/src/slices/frontleadone/index.js
Geoff-Ford: I think I remember this being discussed and that was the agreed way to handle that scenario i.e. render a blank space or an empty ITE

Geoff-Ford (approved):
Looks good. Just that consideration mentioned above regarding showing blank ITE in the absence of data rather than returning `null` for the whole front.

packages/edition-slices/src/slices/frontleadone/index.js
aronshamash: I guess it is possible there to not have `InTheNewsSlice` if there was a problem with publishing or  bug in TPA. I can put the check for `items` in the `InTheEdition` component instead

PR #194, opened 2020-10-20
fix: Lead one styling issue

aronshamash:
- Fixes styling issues on `LeadOneFront`

### DIFF
--- a/packages/edition-slices/src/slices/frontleadone/index.js
+++ b/packages/edition-slices/src/slices/frontleadone/index.js
@@ -38,8 +38,8 @@ function renderMedium(props, breakpoint, orientation) {
 }
 
 const FrontLeadOne = (props) => {
-  const renderSlice = (breakpoint, orientation) =>
-    renderMedium(props, breakpoint, orientation);
+  const renderSlice = (_breakpoint, orientation) =>
+    renderMedium(props, orientation);
 
   return (
     <ResponsiveSlice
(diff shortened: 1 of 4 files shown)


Geoff-Ford (approved):

packages/edition-slices/src/slices/frontleadone/index.js
```diff
@@ -38,8 +38,8 @@ function renderMedium(props, breakpoint, orientation) {
 }
 
 const FrontLeadOne = (props) => {
-  const renderSlice = (breakpoint, orientation) =>
-    renderMedium(props, breakpoint, orientation);
+  const renderSlice = (_breakpoint, orientation) =>
+    renderMedium(props, orientation);
```
c-j-j, line 42: Not sure how this fixes it... but `renderMedium` takes in 3 args, with the second being `breakpoint`.

aronshamash: sorry missed some other changes to remove breakpoint from tile. see next commit

c-j-j: Having said that, `breakpoint` isn't needed for frontleadone, so can be removed

Geoff-Ford: You are right, it is still an argument in `renderMedium` and also passed to one of the tiles

c-j-j (approved):

PR #199, opened 2020-10-20
chore: refactor inline

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-5647

The current approach to inline elements and the absolute positioning of any content around them is very slow and also causes other issues (such as not displaying other inline elements further down the page).

This PR takes a different approach but is at this stage **only applied to inline adverts** aka MPUs. But the logic here could be extrapolated further to be used for inline images and dropcaps.

Basically, if an inline ad is required, at the `body-utils` stage we grab a bunch of paragraphs next to the ad position and pass those along with the ad. The InlineAd then measures this content to see how much of it fits alongside the ad. Any paragraphs, or split paragraphs that don't fit are then rendered as normal below the ad. All content after this, that **wasn't** passed to InlineAd to be measured, will then be rendered below.

Another change required to allow for the above approach was to change `ArticleSkeleton` to return a regular `ScrollView` rather than a `FlatList`. As we are now pre-rendering two articles either side of the current article, this should still be a lot faster that before. Initial renders of the first article page would have been slower due to losing the optimisations you get with `FlatList` however, the inline logic is so much faster that we should still see faster renders on tablet.


packages/ad/src/inline-ad/measure/MeasureInlineContent.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { ContentParameters, Line, SkeletonProps } from "../types";
+
+interface Props {
+  content: ParagraphContent[];
+  contentParameters: ContentParameters;
+  renderMeasuredContents: (contentMeasurements: Measurements) => any;
+  skeletonProps: SkeletonProps;
+}
+
+const allContentMeasured = (
```
c-j-j, line 17: There's only a few differences here between this and the column measuring logic. It would be good when we have more time to try and generalise the two into one. 

Maybe something to look into when we're going to refactor inline images/dropcap to use the same logic.

packages/ad/src/inline-ad/utils/chunkInlineContent.ts
```diff
@@ -0,0 +1,96 @@
...
+      const actualLineToSplit = Math.ceil(provisionalLineToSplit);
+
+      const totalLinesHeightAdjustment = contentLineHeight * actualLineToSplit;
+
+      const adjustmentHeightDiffernce =
+        currentParagraphHeight - totalLinesHeightAdjustment;
+
+      // final adjustment last paragraph lines fit without bottom padding
+      const paddingAdjustment =
+        adjustmentHeightDiffernce <= 20 ? adjustmentHeightDiffernce : 0;
```
c-j-j, line 73: Shall we name this magic number something like `const paragraphMarginBottom = 20`?

packages/article-skeleton/src/body-utils.js
```diff
@@ -52,47 +52,104 @@ const setupArticleMpuTestAd = (
...
+        attributes: {
+          slotName,
+        },
+        children: [],
+      },
+      ...contentAfterControlGroupAd,
+    ];
+  }
+
+  let inlineContentEndIndex = adSlotIndex + 7;
```
c-j-j, line 78: A name for this 7 would be helpful... `const numberOfCandidateParagraphsToInline = 7`?

packages/article-skeleton/src/body-utils.js
```diff
@@ -52,47 +52,104 @@ const setupArticleMpuTestAd = (
...
+
+  let inlineContent = contentWithoutAdSlot.slice(
```
c-j-j, line 80: These `let`'s can be `const`s instead

packages/ad/src/inline-ad/measure/MeasureInlineContent.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { useReducer } from "react";
```
c-j-j, line 1: Is the long term plan for this inline logic to be moved out of `ad/src/inline-ad/measure` into something more high level (for example into something like `packages/inline`) when we refactor inline images/dropcaps?

c-j-j (approved):
Nicely done 👏  only a few minor comments

packages/ad/src/inline-ad/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+  }, []);
+
+  const renderItem = (inline: boolean) => (
+    item: ParagraphContent,
+    index: number,
+  ) => Child({ item, index }, inline);
+
+  const { lineHeight } = defaultFont;
+
+  const adHeaderHeight = 20;
```
aronshamash, line 72: maybe could use multiples of `spacing` ?

packages/ad/src/inline-ad/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { View } from "react-native";
+
+import { ParagraphContent } from "@times-components-native/types";
+import { tabletWidth } from "@times-components-native/styleguide";
+import { render } from "@times-components-native/markup-forest";
+import {
+  renderers,
+  Gutter,
+  ErrorBoundary,
+} from "@times-components-native/article-skeleton";
```
aronshamash, line 11: do you think it is weird to have a low level component dependant on `article-skeleton` ? Not sure ..

packages/ad/src/inline-ad/utils/chunkInlineContent.ts
```diff
@@ -0,0 +1,96 @@
+import { ChunkContents, ContentParameters } from "../types";
+import { splitParagraphContentByLine } from "@times-components-native/utils/src/splitParagraphContentByLine";
```
aronshamash, line 2: we really need to get rid of those `src` dirs but obviously for a different PR.

packages/article-skeleton/src/body-utils.js
```diff
@@ -17,14 +17,14 @@ export const collapsed = (isTablet, content) =>
...
-            node.attributes?.display === "inline") ||
+          (node.name === "image" && node.attributes?.display === "inline") ||
           node.name === "pullQuote"
         ) {
           // forwards
           let i;
           let children = [node];
-          for (i = 0; i < acc.length; i += 1) {
+          const maxNodesToInline = Math.min(4, acc.length);
+          for (i = 0; i < maxNodesToInline; i += 1) {
```
aronshamash, line 27: stick the `4` in const

packages/article-skeleton/src/body-utils.js
```diff
@@ -17,14 +17,14 @@ export const collapsed = (isTablet, content) =>
...
-          ((node.name === "image" || node.name === "ad") &&
-            node.attributes?.display === "inline") ||
+          (node.name === "image" && node.attributes?.display === "inline") ||
           node.name === "pullQuote"
         ) {
           // forwards
           let i;
           let children = [node];
-          for (i = 0; i < acc.length; i += 1) {
+          const maxNodesToInline = Math.min(4, acc.length);
```
aronshamash, line 26: stick the `4` in const

packages/article-skeleton/src/body-utils.js
aronshamash: surprise eslint didn't catch that

packages/types/measurements.ts
```diff
@@ -0,0 +1,12 @@
+type ContentsLine = {
+  text: string;
+};
+
+export interface Measurements {
+  contents: {
+    lines: { [key: string]: ContentsLine[] };
+    heights: { [key: string]: number };
+  };
+  bylineHeight?: number | null;
```
aronshamash, line 10: think putting in a `byline` object nicer ?

aronshamash (commented):
great work @Geoff-Ford !

packages/ad/src/inline-ad/measure/MeasureInlineContent.tsx
Geoff-Ford: Yes exactly. But as you say, when we refactor for images and dropcaps as it may change further.

packages/article-skeleton/src/body-utils.js
Geoff-Ford: eslint didn't catch it because it needs to be a `let`. It can be overridden below if there are non paragraph content items.

packages/ad/src/inline-ad/measure/MeasureInlineContent.tsx
Geoff-Ford: Absolutely!

packages/ad/src/inline-ad/index.tsx
Geoff-Ford: It isn't dependant on `article-skeleton` per se, it is some of the components and the renderers that are required here. `ArticleSkeleton` is somewhat of a crowded component. We could look at breaking some of these things out in a later PR.

packages/ad/src/inline-ad/utils/chunkInlineContent.ts
Geoff-Ford: Agreed, will be a fair bit of work to ensure exports all work still but yep this should still be on our list. However, in this case you would normally not see the `src/etc` part of the path as you would be importing from the root of that package. The issue is that you can't mix `ts` and `tsx` files with `js` files in your root exports. They don't work properly. Thus having to import TypeScript files directly.

packages/article-skeleton/src/body-utils.js
Geoff-Ford: This could actually go. It was an optimisation for the ads. I suppose it still might make inline images a little faster. 🤔  ... okay, I'll keep it and add a const.

packages/types/measurements.ts
Geoff-Ford: This is just a merged type between the `ArticleColumn`'s measurements and the new measurements. So this type works in both places. What do you mean by an `object` for `byline`? The value is a `number` if it is set.

aronshamash: I meant 
```
byline: {
   height: 
}
```

packages/ad/src/inline-ad/index.tsx
aronshamash: yes I think that is more the issue

aronshamash (approved):

PR #202, opened 2020-10-21
fix: Align ITE item CTA text to bottom

aronshamash:
- align ITE item CTA text to bottom

### Before 
Screenshot 2020-10-21 at 14 52 19

### After
Screenshot 2020-10-21 at 15 02 29

### DIFF
--- a/packages/in-todays-edition/styles/index.js
+++ b/packages/in-todays-edition/styles/index.js
@@ -37,7 +37,6 @@ const sharedStyles = {
     fontFamily: fonts.supporting,
     color: colours.functional.red,
     textDecorationLine: "none",
-    marginBottom: spacing(3),
   },
   itemCTAIconContainer: {
     marginLeft: spacing(1),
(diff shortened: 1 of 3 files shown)


packages/in-todays-edition/styles/index.js
```diff
@@ -71,6 +70,16 @@ const sharedPortraitStyles = {
   itemLast: {
     marginRight: 0,
   },
+  itemCTA: {
+    flexDirection: "row",
+    flex: 1,
+    alignItems: "flex-end",
+    marginBottom: -2,
```
c-j-j, line 77: what's the reason for these 2px margins?

c-j-j (approved):

packages/in-todays-edition/styles/index.js
```diff
@@ -71,6 +70,16 @@ const sharedPortraitStyles = {
...
+  },
+  itemCTAIconContainer: {
+    ...sharedStyles.itemCTAIconContainer,
+    marginBottom: 2,
```
aronshamash, line 81: this one is equivalent to the marginTop: 2 on landscape needed to align CTA text and icon

packages/in-todays-edition/styles/index.js
aronshamash: this one is strange but with `alignItems` set to flex-end it still wasn't as far down as previously. It was 2px out.

PR #205, opened 2020-10-22
fix: Adds correct margins to headlines and optional straplines

c-j-j:
Simulator Screen Shot - iPad mini 4 - 2020-10-22 at 15 08 22
Simulator Screen Shot - iPad mini 4 - 2020-10-22 at 15 08 16


aronshamash (approved):
looks good

PR #206, opened 2020-10-23
fix: Only hyphenates justified content

c-j-j:
This PR changes the front logic to only hyphenate justified content. 

Furthermore, previously we were only left aligning text in the support tile when it was a maincomment article. As justified content looks bad in a narrow column, we're now always left-aligning text in the support tile.

Also - I fixed the summary text size on the first lead one and one tile in landscape, as it didn't match the support tile.

image

### DIFF
--- a/packages/edition-slices/src/tiles/tile-h-front/index.js
+++ b/packages/edition-slices/src/tiles/tile-h-front/index.js
@@ -27,8 +27,8 @@ const TileHFront = ({ onPress, tile, orientation }) => {
         summaryLineHeight={styles.summary.lineHeight}
         bylines={article.bylines}
         bylineMarginBottom={styles.bylineMarginBottom}
+        justified={false}
         tile={tile}
-        template={article.template}
       />
     </TileLink>
   );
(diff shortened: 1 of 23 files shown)


packages/edition-slices/src/tiles/tile-h-front/index.js
```diff
@@ -27,8 +27,8 @@ const TileHFront = ({ onPress, tile, orientation }) => {
         summaryLineHeight={styles.summary.lineHeight}
         bylines={article.bylines}
         bylineMarginBottom={styles.bylineMarginBottom}
+        justified={false}
```
aronshamash, line 30: do we need to explicitly pass `false`? shouldn't it be the default ?

packages/front-page/__tests__/front-article-summary-content.test.js
```diff
@@ -31,20 +31,25 @@ const ast = [
...
-    style={{ backgoundColor: "red" }}
-    summary={ast}
-  />
-);
 
+const props = {
+  bylines: bylines,
+  whiteSpaceHeight: 20,
+  justified: false,
+  linesOfTeaserToRender: 1,
```
aronshamash, line 39: `teaserTextLineCount` ?

c-j-j: this prop actually isn't used anymore by FrontArticleSummaryContent so I will remove altogether.

packages/edition-slices/src/tiles/tile-h-front/index.js
c-j-j: yes can do that

aronshamash (approved):

PR #214, opened 2020-10-26
fix: Aligns tile summary content to top of tile

c-j-j:
For the larger iPads, we were aligning text to the bottom of the tile due to the absolute positioning that we introduced.

image

To fix this, we're assigning the minHeight of this container to be the height that is available. Also, I've set the marginBottom of the lead image to 0, and instead am setting paddingTop on the summary container to ensure there's always the right padding between image and headline.

image
image

image

### DIFF
--- a/packages/front-page/front-tile-summary.tsx
+++ b/packages/front-page/front-tile-summary.tsx
@@ -109,15 +109,15 @@ const FrontTileSummary: React.FC<Props> = (props) => {
     straplineHeight !== undefined &&
     bylineHeight !== undefined;
 
-  const TileSummaryContainer: React.FC<{ hidden: boolean }> = ({
-    children,
-    hidden,
-  }) => (
+  const TileSummaryContainer: React.FC<{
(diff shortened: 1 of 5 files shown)


aronshamash (approved):
Great looks good.

packages/front-page/front-tile-summary.tsx
```diff
@@ -109,15 +109,15 @@ const FrontTileSummary: React.FC<Props> = (props) => {
...
+  const TileSummaryContainer: React.FC<{
+    hidden: boolean;
+    minHeight?: number;
+  }> = ({ children, hidden, minHeight }) => (
     <View
       style={[
         props.containerStyle,
         styles.container,
-        { opacity: hidden ? 0 : 1 },
+        { minHeight, opacity: hidden ? 0 : 1 },
```
Geoff-Ford, line 120: As it is optional, should this check first or does it matter if the value is `undefined`?

PR #219, opened 2020-10-26
fix: Three deck headline and col count

aronshamash:
- set col count to 3 on both orientations
- absolutely position headline

### Before
Screenshot 2020-10-26 at 17 15 39

### After
Screenshot 2020-10-26 at 17 21 06

### Showing body content with short headline and no standfirst (edge case)
Screenshot 2020-10-26 at 17 06 08

### DIFF
--- a/packages/edition-slices/src/tiles/tile-f-front/index.js
+++ b/packages/edition-slices/src/tiles/tile-f-front/index.js
@@ -14,8 +14,6 @@ import { getDimensions } from "@times-components-native/utils";
 
 const TileFFront = ({ onPress, tile, orientation }) => {
   const { width: windowWidth, height: windowHeight } = getDimensions();
-  const isLandscape = orientation === "landscape";
-  const columnCount = isLandscape ? 1 : 3;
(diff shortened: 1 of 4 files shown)


packages/edition-slices/src/tiles/tile-f-front/index.js
```diff
@@ -48,7 +47,7 @@ const TileFFront = ({ onPress, tile, orientation }) => {
         justified={true}
         straplineMarginBottom={styles.straplineMarginBottom}
         tile={tile}
-        columnCount={columnCount}
+        columnCount={3}
```
c-j-j, line 50: This is the first time we're rendering columns in a relatively small space - so I think we could render columns with 2 lines of text if there's space for it - is that going to be sufficient?

c-j-j (approved):

packages/edition-slices/src/tiles/tile-f-front/index.js
aronshamash: yes I'll post a screenshot if I can publish to staging.

PR #222, opened 2020-10-28
fix: fixes columns on lead1 landscape

c-j-j:
**Problem**
image

When rendering a front tile, we were previously measuring how much space is available twice. Firstly, we measured if we can fit headline+strapline+byline+content. Secondly, we measured how much room is available for columns - so we can chunk the content as required.

As we are now absolutely positioning the tile summary to allow it to grow, we're seeing issues with the second measuring phase being done in an absolutely positioned view. The measured heights are coming back all wrong - which is why we're seeing this large column being rendered.

**Solution**
We don't really need the second measuring step. We can use the info captured from the first step, and infer how much space is available for teaser text. This fixes the issue on lead1 landscape.


aronshamash (approved):
looks good.

PR #224, opened 2020-10-29
feat: Section Ads

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-3682


packages/section/__tests__/unit/utils.test.js
```diff
@@ -63,7 +64,10 @@ describe("composeSliceBuilders", () => {
       { id: "l", name: "DailyUniversalRegister" },
     ];
 
-    const buildData = composeSliceBuilders(builderOne, builderTwo);
+    const buildData = prepareSlicesForRender(false, variants)(
+      builderOne,
```
aronshamash, line 68: I take it all these `builders` will go ?

packages/ad/src/utils/ad-init.js
```diff
@@ -231,6 +231,11 @@ export default ({ el, data, platform, eventCallback, window }) => {
               const testGroup = ["a", "b", "c"].indexOf(slotName.slice(-1)) + 1;
               slot.setTargeting("testgroup", testGroup.toString());
             }
+            if (/^native-section-ad-/.test(slotName)) {
+              const testGroup = ["a", "b", "c"].indexOf(slotName.slice(-1)) + 1;
```
aronshamash, line 235: can this be DRYed up ?

packages/ad/src/utils/generate-config.js
```diff
@@ -43,6 +43,8 @@ const sizeMap = {
   "native-inline-ad-a": sizes.native,
   "native-inline-ad-b": sizes.nativeB,
   "native-inline-ad-c": sizes.nativeC,
+  "native-section-ad-b": sizes.sectionB,
```
aronshamash, line 46: why no `sectionA`?

packages/pages/src/section/section.js
```diff
@@ -131,9 +134,18 @@ class SectionPage extends Component {
...
+    const { error, isLoading, publicationName, variants } = this.props;
     const { recentlyOpenedPuzzleCount, savedArticles, section } = this.state;
 
+    const adConfig =
+      isLoading || error
+        ? {}
+        : adTargetConfig({
+            // adTestMode,
+            // article,
+            sectionName: section.name,
```
aronshamash, line 146: do we need a default or will this always be populated ?

packages/pages/src/section/ad-targeting-config.js
```diff
@@ -0,0 +1,32 @@
+import { NativeModules, Platform } from "react-native";
+
+const config = NativeModules.ReactConfig;
+
+export default ({ adTestMode = null, sectionName }) => ({
```
aronshamash, line 5: do you think it is worth having a base ad config with all the common stuff in ?

packages/ad/src/utils/generate-config.js
aronshamash: I worked it out, there is no need for it!

packages/variant-testing/src/variant-testing.tsx
```diff
@@ -1,42 +1,53 @@
 import React, { createContext, useContext, ReactNode } from "react";
 import PropTypes from "prop-types";
 
+import { useResponsiveContext } from "@times-components-native/responsive";
+
 export const VariantTestingContext = createContext({});
 
+const validateVariant = (variant: string) => {
```
aronshamash, line 8: yes better.

packages/ad/src/utils/ad-init.js
c-j-j: You may have made this comment pre-DRYing - pushed a change to extract that into one regex

packages/pages/src/section/ad-targeting-config.js
c-j-j: Yes that will make sense - although  they are due to audit this config for section and articles, so exactly config should be sent for article vs section is unclear right now

packages/pages/src/section/section.js
c-j-j: I had assumed section.name is always set - can we assume that's the case, and change if we see undefined section names?

packages/ad/src/utils/ad-init.js
aronshamash: I did

packages/pages/src/section/section.js
aronshamash: yes. we are doing `sectionName.toLowerCase()` so that would break if sectionName is `undefined`

aronshamash (approved):

packages/section/__tests__/unit/utils.test.js
c-j-j: I think these tests can probably all go - they are trying to test the piping logic which is now done by the `pipe` util and its tested separately. They don't check any of the data transformation. 

I think I'll add some tests that check prepareSlicesForRender with real input/output data and snapshot the result - as t he internal bits of logic are also tested separately

PR #230, opened 2020-11-02
fix(TNLT-5484): fixes android orientation issues

c-j-j:
**Problem**
This PR fixes the issues surrounding android not responding to orientation changes correctly. There are two issues to solve:
1. The RN Dimensions event listener doesn't trigger on orientation changes
2. Whenever we call `Dimensions.get('window')` to fetch the current width/height,  we only ever see the width/height when the app first loaded. This is problematic whenever we change sections/articles, as we are using `Dimension.get('window')` to fetch the initial dimensions before subscribing to the event listener.

**Solution**
Android now provides the screen height/width with the initial props for the section/article pages. Android also provides a custom event listener that RN can use to subscribe to orientation changes. 

In order to propagate these height/width values to everywhere required, we are adding these to the Responsive context. This PR changes most calls to `getDimensions` to `useResponsiveContext` instead.

FYI On iOS, we are still using the RN Dimensions API for everything as it works as expected.

**Important changes:**
The Responsive Context is now given the correct initial values. Previously, we only had a subset of the values defined (isTablet/orientation) - and these were set to arbitrary values. Now they are all present and correct.

The Responsive package is now written in typescript and refactored to a functional component.

https://github.com/newsuk/times-components-native/pull/230/files#diff-a212c2c6ad3b37b133ab03d16cf464d4b679bd96d89a9a1486b4848af9b87ac3

https://github.com/newsuk/times-components-native/pull/230/files#diff-1abb782baf65c362f44466e88e3c37f8e2797f7e749f53adab90faf4bbd86b00

**Why are so many unit tests/snapshots changed in this PR?**
Many of the unit tests were dependent on the Responsive context's initial values, which were hardcoded to `isTablet=false; orientation="landscape"`. In order to ensure that mobile/tablet snapshots were actually for mobile/tablet, I had to explicitly provide the correct responsive context for the various tests. Some of the previous tablet snapshots were actually for mobile, as the `setIsTablet(true)` call doesn't work.

**What still doesn't work?**
We only get the initial height/width values for the section/article pages. We still need the following pages to receive them:
- [ ] Related articles (dropcap fails on this page without correct width)
- [ ] Topics (no noticeable UI issues)
- [ ] Author (no noticeable UI issues)

Also, some logic still has a dependency on the RN dimensions API, like the full screen image modal. I haven't changed these, as there's a fair bit of complexity there that I haven't had time to go through. Hopefully, RN will push fixes for https://github.com/facebook/react-native/issues/29451 and we won't need to refactor these parts.


packages/utils/src/dimensions-util.js
```diff
@@ -7,9 +18,28 @@ export const getDimensions = (
...
+    const wrappedHandler = (data) => {
+      return handler({
+        window: {
+          width: data.displayWidth,
+          height: data.displayHeight,
+          fontScale: data.fontScale,
+        },
+      });
+    };
+    eventEmitter.addListener("OrientationChanged", wrappedHandler);
```
aronshamash, line 31: very minor but don't we have convention of lower camel case for events eg in https://github.com/newsuk/times-components-native/blob/master/packages/pages/src/section/section.js

packages/utils/src/dimensions-util.js
```diff
@@ -7,9 +18,28 @@ export const getDimensions = (
...
+    return wrappedHandler;
+  } else {
+    Dimensions.addEventListener(type, handler);
+    return handler;
+  }
 };
 
 export const removeDimensionsListener = (type, handler) => {
-  Dimensions.removeEventListener(type, handler);
+  if (Platform.OS === "android") {
```
aronshamash, line 40: could be ternary no ?

packages/utils/src/dimensions-util.js
c-j-j: I can ask Christian to change

packages/test-utils/src/responsiveContextUtil.tsx
```diff
@@ -0,0 +1,15 @@
+import { ResponsiveContext } from "@times-components-native/responsive";
+import { calculateResponsiveContext } from "@times-components-native/responsive/src/calculateResponsiveContext";
+import React from "react";
+
+export const withTabletContext = (WrappedComponent: any) => (
+  <ResponsiveContext.Provider value={calculateResponsiveContext(1000, 400, 1)}>
```
aronshamash, line 6: worth DRYing this up ?

packages/slice-layout/__tests__/ios/__snapshots__/front-l1-with-style.ios.test.js.snap
```diff
@@ -117,7 +117,7 @@ exports[`6. front lead one - landscape - 1080 1`] = `
       Array [
         Object {
           "style": Object {
-            "width": "75.5031779661017%",
+            "width": "75.47500000000001%",
```
aronshamash, line 120: do you know reason for this change ?

packages/utils/src/dimensions-util.js
c-j-j: Not sure if ternary is appropriate here as it's not an expression/there's no return value. I'm sure eslint would usually complain if you try and do `foo ? bar() : bat()`

aronshamash (approved):
Great work @c-j-j.

packages/slice-layout/__tests__/ios/__snapshots__/front-l1-with-style.ios.test.js.snap
c-j-j: It's because the previous percentage was wrong. In the `columnToPercentage` function - there's a call to get the window width `const windowWidth = Dimensions.get("window").width;` - which was always returning 750 in the test. Now it's passed in and not mocked

packages/responsive/src/context.ts
```diff
@@ -0,0 +1,19 @@
...
+  isTablet: boolean;
+  windowWidth: number;
+  windowHeight: number;
+  orientation: string;
+};
+
+const { width, height, fontScale } = getDimensions();
+
+export default createContext<ContextType>(
+  calculateResponsiveContext(width, height, fontScale),
```
Geoff-Ford, line 18: Not a biggie, but does this need to be a function?

Geoff-Ford: Ignore me. I can see this is re-used.

Geoff-Ford (approved):

packages/test-utils/src/responsiveContextUtil.tsx
c-j-j: I'm not seeing how we could remove the duplication here. What did you have in mind?

packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -357,10 +359,19 @@ export default ({
         caption,
       },
     ) {
+      // eslint-disable-next-line react-hooks/rules-of-hooks
+      const { windowWidth } = useResponsiveContext();
       const aspectRatio = 16 / 9;
+
+      const getWideContentWidth = () => {
+        const contentWidth = isTablet ? tabletWidth : windowWidth;
+        return contentWidth - (isTablet && tabletRowPadding);
```
Geoff-Ford, line 368: Had to check this worked. Must be coerced.

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -45,13 +45,14 @@ const ArticleWithContent = (props) => {
...
   const variants = useVariantTestingContext();
+  const { windowWidth } = useResponsiveContext();
 
   const { id, url, content, template } = data;
 
   const header = useMemo(
     () => (
       <Gutter>
-        <Header width={Math.min(maxWidth, screenWidth())} />
+        <Header width={Math.min(maxWidth, windowWidth)} />
```
Geoff-Ford, line 55: As this is memoized, do you have to update if `windowWidth` changes?

packages/test-utils/src/responsiveContextUtil.tsx
aronshamash: just having a function which takes `width` and returns:
```
  <ResponsiveContext.Provider value={calculateResponsiveContext(width, 400, 1)}>
    {WrappedComponent}
  </ResponsiveContext.Provider>
```

c-j-j: Ah okay, I have changed to this which does look nicer

```
const withContext = (WrappedComponent: any, width: number) => (
  <ResponsiveContext.Provider value={calculateResponsiveContext(width, 400, 1)}>
    {WrappedComponent}
  </ResponsiveContext.Provider>
);

export const withTabletContext = (WrappedComponent: any) =>
  withContext(WrappedComponent, 1000);

export const withMobileContext = (WrappedComponent: any) =>
  withContext(WrappedComponent, 300);

```

packages/article-skeleton/src/article-skeleton.js
c-j-j: Yes, correct. I'll add for now. But would be good to remove this useMemo once all the inline stuff is refactored to the new approach.

packages/article-skeleton/src/article-body/article-body-row.js
c-j-j: Also confused me. Didn't realise `1-false == 1`  whereas `1-undefined == undefined`

PR #234, opened 2020-11-06
fix: Only show bylines in lead comment and cartoon slice

aronshamash:
- Fixes https://nidigitalsolutions.jira.com/browse/TNLT-5680 and https://nidigitalsolutions.jira.com/browse/TNLT-5573
- Currently we can have situation with duplicate author names shown in the
 `CommentLeadAndCartoon` slice as editorial put the author name in the label. To improve the situation we are going to hide the label (even it is populated) and only show the byline (which is almost always populated)
- Adds byline field to the tablet tile as it was also missing
- Fixes layout issue whereby the byline was too close to the author image
- There will be a follow up PR in render to align all platforms

### Before
Screenshot 2020-11-06 at 12 53 56

Screenshot 2020-11-06 at 12 53 39

Screenshot 2020-11-06 at 15 13 18

### After
Screenshot 2020-11-06 at 12 53 31

Screenshot 2020-11-06 at 12 54 10

Screenshot 2020-11-06 at 15 19 03


packages/edition-slices/__tests__/android/__snapshots__/tile-a.test.js.snap
```diff
@@ -20,6 +20,7 @@ exports[`tile a without breakpoint 1`] = `
     >
       <ArticleLabel
         color="#005B8D"
+        hide={false}
```
c-j-j, line 23: Given that `hide` is equal to false most of the time, can it be set as the default prop so we don't need to update all of these snapshots?

c-j-j (approved):

packages/edition-slices/__tests__/android/__snapshots__/tile-a.test.js.snap
aronshamash: yes I know I noticed this but I already set it to default to false : https://github.com/newsuk/times-components-native/pull/234/files#diff-2a015173eab3fade57c3c6b34549ae1cc2211e18a247a1c6554c4fc000d4ac76R118

PR #237, opened 2020-11-10
feat: (TNLT-5482) Commenting tooltip

aronshamash:
- https://nidigitalsolutions.jira.com/browse/TNLT-5482
- Creates a basic `Tooltip` component initially for use only for commenting. No placement of the arrow or positioning is currently available to consumers. This will come in future PRs.
- Uses the `tooltip` prop supplied by native which is of the form:

```
{
  "tooltips": ["commenting"]
}
```
- The array has all the tooltips that need to be shown.
- Tooltips should be tablet only
- The tooltip calls the `onTooltipPresented` callback when the tooltip is presented to the user.

### iPad Mini 4
Nov-11-2020 13-16-13

### iPhone 8
Screenshot 2020-11-10 at 16 42 31

Outstanding:
- [x] Unit testing
- [x] Get the close button supplied by UXD working
- [x] Make work for all articles templates


ios/Podfile.lock
```diff
@@ -464,7 +464,7 @@ SPEC CHECKSUMS:
   boost-for-react-native: 39c7adb57c4e60d6c5479dd8623128eb5b3f0f2c
   CocoaAsyncSocket: 694058e7c0ed05a9e217d1b3c7ded962f4180845
   CocoaLibEvent: 2fab71b8bd46dd33ddb959f7928ec5909f838e3f
-  DoubleConversion: cde416483dac037923206447da6e1454df403714
+  DoubleConversion: 5805e889d232975c086db112ece9ed034df7a0b2
```
c-j-j, line 467: This shouldn't have changed - did you mean to commit this?

packages/article-comments/src/comments.js
```diff
@@ -43,8 +40,41 @@ const Comments = ({
...
+  );
+
+  const renderButtonWithTooltip = () => (
+    <Tooltip
+      content={
+        <Text>Tap to read comments and join in with the conversation</Text>
+      }
+      onTooltipPresented={onTooltipPresented}
+      type={tooltipType}
+    >
```
c-j-j, line 52: Just wondering if we could avoid having to do this everywhere we want to show a tooltip:

```
  const showCommentButton =
    tooltips?.includes(tooltipType) && isTablet
      ? renderButtonWithTooltip
      : renderButton;
```

Could we instead have the Tooltip component do that internally?

packages/article-comment-tablet/src/article-comment-tablet.js
```diff
@@ -148,6 +150,7 @@ class ArticlePage extends Component {
                     onImagePress={onImagePress}
                     onLinkPress={onLinkPress}
                     onRelatedArticlePress={onRelatedArticlePress}
+                    onTooltipPresented={onTooltipPresented}
```
c-j-j, line 153: One day it'll be nice to move all these native callbacks into its own Context (named something like `NativePropContext`) - to remove having all this prop drilling on these components.

packages/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useEffect, useState } from "react";
+import { Animated, Text, TouchableOpacity, View } from "react-native";
+
+import { colours } from "@times-components-native/styleguide";
+import { IconClose } from "@times-components-native/icons";
+import styles from "./styles";
+
+interface Props {
+  content: any;
+  children: any;
```
c-j-j, line 10: Think you can remove `children` here as React.FC does that

packages/tooltip/tooltip.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useEffect, useState } from "react";
+import { Animated, Text, TouchableOpacity, View } from "react-native";
+
+import { colours } from "@times-components-native/styleguide";
+import { IconClose } from "@times-components-native/icons";
+import styles from "./styles";
+
+interface Props {
+  content: any;
```
c-j-j, line 9: `content: string`?

ios/Podfile.lock
aronshamash: no always happens for me. I'll revert now

packages/article-comments/src/comments.js
aronshamash: yes I was thinking that too. makes sense

packages/article-comment-tablet/src/article-comment-tablet.js
aronshamash: yep. again I was thinking same thing when passing them down. could be something for following PRs.

packages/tooltip/tooltip.tsx
aronshamash: didn't know that. thanks

packages/article-comment-tablet/src/article-comment-tablet.js
c-j-j: Yes - really it'll be its own ticket, as you'd want to pass all the props down, and test everywhere to make sure they all work

packages/article-comments/src/comments.js
aronshamash: I have moved logic to the tooltip component now.

packages/tooltip/__tests__/shared.js
```diff
@@ -0,0 +1,117 @@
...
+import TestRenderer from "react-test-renderer";
+import { ResponsiveContext } from "@times-components-native/responsive";
+import { delay } from "@times-components-native/test-utils";
+import { shallow } from "enzyme";
+import { TouchableOpacity, Animated } from "react-native";
+import React from "react";
+import { Text } from "react-native";
+import "./serializers-with-all-styles";
+
+export const withTabletContext = (WrappedComponent, isTablet = true) => (
```
c-j-j, line 11: the test-utils package now has a withTabletContext HOC that can you use instead now

c-j-j (approved):
Nice work

packages/tooltip/__tests__/shared.js
aronshamash: thanks I'll update the PR

packages/article-comment-tablet/src/article-prop-types/article-prop-types.base.js
```diff
@@ -13,13 +13,15 @@ const articlePagePropTypes = {
   isLoading: PropTypes.bool,
   onImagePress: PropTypes.func,
   receiveChildList: PropTypes.func.isRequired,
+  tooltips: PropTypes.array,
```
Geoff-Ford, line 16: Does `onTooltipPresented` also need to be added to the propTypes?

Geoff-Ford: Oh, I see. You add it in later.

packages/tooltip/styles/index.js
```diff
@@ -0,0 +1,72 @@
+import { StyleSheet } from "react-native";
+import { fonts, spacing, colours } from "@times-components-native/styleguide";
+
+const styles = StyleSheet.create({
+  container: {
+    padding: spacing(3),
+    width: 256,
+    backgroundColor: colours.functional.tooltip,
+    position: "absolute",
+    bottom: -5,
```
Geoff-Ford, line 10: Is this coincidentally -5 or can this be `-spacing(1)`?

packages/tooltip/styles/index.js
```diff
@@ -0,0 +1,72 @@
...
+    left: -128,
```
Geoff-Ford, line 11: Another really minor thing if you want it. Maybe make the container width a const then you can set the `left` to `-(containerWidth/2)`? I assume this is to centre it?

Geoff-Ford (approved):
Looks really good. A couple of very minor points which 🤷‍♂️

packages/tooltip/styles/index.js
aronshamash: it will change in the next pr but made the change

packages/article-comment-tablet/src/article-prop-types/article-prop-types.base.js
aronshamash: Yes but there are in a different file. Here is an example: https://github.com/newsuk/times-components-native/pull/237/files#diff-395d00f44859761a216336fe783191e59cccd442ba9538d0e125fdc1898041d0R16

packages/tooltip/styles/index.js
aronshamash: yes I am going to need to do this when we need tooltip width to be passed in as a prop

PR #242, opened 2020-11-12
chore: pullQuote and app context changes

Geoff-Ford:
Moved `pullQuote` `Context` and `View` wrapper from article-body-row into the `PullQuote` component.
Also removed `user` from the app context object as not used anymore and removed all outdated references to it.


packages/pull-quote/__tests__/shared-with-style.base.js
```diff
@@ -1,17 +1,24 @@
...
 
+jest.mock("@times-components-native/context");
+
 const content = "Some content";
 const caption = "A caption";
 const twitter = "@twitter";
 
 export default (renderComponent) => {
   it("different colours", () => {
+    useAppContext.mockImplementation(() => ({
```
c-j-j, line 14: Just out of interest, are you mocking this because jest is having difficulties with the hooks? Another approach is to wrap the test in a context provider as I did with the mobile/tablet contexts.

c-j-j (approved):

packages/pull-quote/__tests__/shared-with-style.base.js
Geoff-Ford: The code I am testing is using the new hook so just thought it made sense to just mock that as it is easy. Also the context is different in the different tests i.e. shared tests and theme tests.

PR #244, opened 2020-11-15
feat(TNLT-4872): Update newsletter puff tracking

themischatzis:
### Description 
In this PR we fix the `onPress` and `oneView` tracking by adding the `trackingName` to the NewsletterPuff components.

### Jira Ticket
[TNLT-4872](https://nidigitalsolutions.jira.com/browse/TNLT-4872)

### Checklist
- [x] Does it have automated tests?
- [x] Is there another PR linked to this issue (if so please list)?
 PR [#2654](https://github.com/newsuk/times-components/pull/2654) in `times-components`

### DIFF
--- a/packages/tracking/src/tracking-context.js
+++ b/packages/tracking/src/tracking-context.js
@@ -82,7 +82,7 @@ const withTrackingContext = (
     attemptTrackPageEvent(props) {
       if (
         isDataReady(props) &&
-        this.isRootTrackingContext() &&
+        (this.isRootTrackingContext() || props.newsletterPuffName) &&
         this.pageEventTriggered === false
       ) {
         this.pageEventTriggered = true;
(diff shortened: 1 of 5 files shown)


packages/tracking/src/tracking-context.js
```diff
@@ -82,7 +82,7 @@ const withTrackingContext = (
     attemptTrackPageEvent(props) {
       if (
         isDataReady(props) &&
-        this.isRootTrackingContext() &&
+        (this.isRootTrackingContext() || props.newsletterPuffName) &&
```
c-j-j, line 85: This tracking-context currently doesn't have any business domain in so far - I think it'll make more sense to keep it general and create a new prop called something like `enforceTracking` instead of checking `props.newsletterPuffName`.

c-j-j (approved):

PR #245, opened 2020-11-16
feat: Topic tooltips

aronshamash:
- https://nidigitalsolutions.jira.com/browse/TNLT-3502
- Adding new props of `offsetY`, `width` and and `onClose` callback to `tooltip` component
- The next PR for author profile tooltip (which has additional requirements), we can refactor styles to use a `generateStyles` function to make the styling a little cleaner.

### Comment tablet
Nov-16-2020 16-33-29

### Main standard
Nov-16-2020 16-33-55


packages/article-topics/src/article-topic.js
```diff
@@ -1,40 +1,76 @@
...
-            onPress={(e) => onPress(e, { name, slug })}
-            url={makeTopicUrl({ slug })}
-          >
-            <View style={styles.container}>
-              <Text
-                accessibilityComponentType="button"
-                accessibilityRole="button"
-                accessibilityTraits="button"
-                style={[styles.text, fontSizeStyle, lineHeightStyle]}
+    showTooltip && (
```
c-j-j, line 31: If `showTooltip` is false - would that result in this articleTopic component being null?

aronshamash: ha good spot. that is a bug. I've sorted it now and added tests for style helper

c-j-j (approved):

packages/article-topics/src/article-topic.js
```diff
@@ -32,9 +55,26 @@ const ArticleTopic = ({ fontSize, lineHeight, name, onPress, slug }) => {
       )}
     </Context.Consumer>
   );
+
+  const articleTopicTooltip = (
```
Geoff-Ford, line 59: `articleTopicWithTooltip`?

Geoff-Ford (approved):

PR #246, opened 2020-11-17
chore: Refactor inline images and pull quotes

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-5785

* Refactored inline images and pull quotes to use the new approach to inline content already applied to ads.
* Also unify the inline ad code into the new inline component and remove the previous ad specific code.

Also fixes https://nidigitalsolutions.jira.com/browse/TNLT-4916


packages/inline-content/src/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+const InlineContent = (props: InlineContentProps) => {
+  const {
+    defaultFont,
+    inlineContent,
+    narrowContent,
+    originalName,
+    skeletonProps,
+  } = props;
+  const { windowWidth } = useResponsiveContext();
+  const { lineHeight } = defaultFont;
```
c-j-j, line 36: I assume this lineHeight will change when changing both device font settings and app font settings?

packages/inline-content/src/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+  let adContainerHeight: number;
+
+  if (isAd) {
+    const { height: adHeight, width: adWidth } = props;
+    const adHeaderHeight = spacing(4);
+    const adHorizontalSpacing = 21;
+    const adMarginBottom = spacing(2);
+    adContainerHeight = adHeight + adHeaderHeight;
+    inlineContentHeight = adContainerHeight + adMarginBottom;
+    inlineItemWidth = adWidth + adHorizontalSpacing;
```
c-j-j, line 56: just out of interest, what was the reason why you didn't render an inline ad to capture its dimensions (similar to how you've done this for inline images/pull quotes/etc)? Is it because we don't want to try and fetch an actual ad from the ad server as part of doing this?

c-j-j (approved):
👏

packages/inline-content/src/index.tsx
Geoff-Ford: Two reasons. Firstly we already know what size the ad will be as this comes from the config/variant and the additional padding etc. So as these are fixed values there is no benefit in measuring them. And secondly, yes, we don't want to create a rendered impression on the ad when it hasn't actually been seen.

packages/inline-content/src/index.tsx
Geoff-Ford: Should do yes. `defaultFont` is created in `article-body-row` and uses the `scale` passing in by native to calculate font sizes via the `styleguide`.

packages/inline-content/__tests__/test-utils/index.ts
```diff
@@ -0,0 +1,25 @@
+import { Line } from "../../src/types";
+import { ParagraphContent } from "@times-components-native/types";
+
+const range = (n: number) => [...Array(n).keys()];
+
+export const createTextWithNumberOfLines = (
+  numberOfLines: number,
```
aronshamash, line 7: could be `lineCount`

packages/inline-content/__tests__/test-utils/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  offset = 0,
+): string =>
+  range(numberOfLines)
+    .map((i) => `line${i + offset}`)
+    .join("");
+
+export const createLinesWithNumberOfLines = (numberOfLines: number): Line[] =>
```
aronshamash, line 14: `createLinesByLineCount` ? no biggie.

aronshamash (approved):
looks very good.

packages/inline-content/__tests__/test-utils/index.ts
Geoff-Ford: I used the same terminology as we already have in `article-columns`. And I guess `numberOfLines` is used because that is the attribute name on `Text`.

packages/inline-content/__tests__/test-utils/index.ts
Geoff-Ford: Same as above comment

PR #256, opened 2020-11-25
feat: (TNLT-5967) Add tooltip for bylines with authors and refactor tooltip component

aronshamash:
- https://nidigitalsolutions.jira.com/browse/TNLT-5967
- Add tooltip for bylines with authors and refactors tooltip component
- Tooltip now takes more props to make it more flexible. 
- The author tooltip is not shown on the in-depth article template due to layout issues relating to zIndex in React Native.
- The commenting tooltip has now been placed to the right of the button.

### iPad and iPhone 8  Comment article template
Screenshot 2020-11-24 at 14 01 17

 
### iPad Mini Main standard article template
Screenshot 2020-11-25 at 14 59 03

### iPad and iPhone 8  Magazine standard article template
Screenshot 2020-11-24 at 14 34 40

### iPad  Indepth article template
Screenshot 2020-11-24 at 14 35 36

### Android  Pixel C
Screenshot 2020-11-25 at 12 54 42


packages/article-byline/src/article-byline-with-links.js
```diff
@@ -32,12 +34,49 @@ const AuthorComponent = ({
...
+    tooltipOffsetY,
+    tooltips,
+  } = props;
+
   const textStyle = centered ? [styles.text, styles.centered] : styles.text;
-  return renderByline(withTrackEvents(AuthorComponent), ast, textStyle, props);
-};
+  const tooltipType = "profile";
+  const showTooltip =
+    !disableTooltip && hasAuthorData(ast) && tooltips.includes(tooltipType);
```
c-j-j, line 50: are we showing tooltips on mobile? if not, I assume disableTootip is set to false?

packages/tooltip/tooltip.tsx
```diff
@@ -61,30 +73,16 @@ const Tooltip: React.FC<Props> = ({
...
-              <Text style={[styles.text, leftAligned && styles.textLeft]}>
-                {content}
-              </Text>
-              <View
-                style={[
-                  styles.arrow,
-                  { left: calculateArrowPosition(alignment, width) },
-                ]}
-              />
+              <Text style={styles.text}>{content}</Text>
```
c-j-j, line 78: Maybe want to set `allowFontScaling` to false here? I assume we wouldn't want giant text in this tooltip if the user has it set on their device?

packages/article-magazine-standard/src/article-magazine-standard.js
```diff
@@ -54,24 +62,22 @@ class ArticleMagazineStandard extends Component {
...
               getImageCrop={getStandardTemplateCrop}
               onImagePress={onImagePress}
               onVideoPress={onVideoPress}
               renderCaption={({ caption }) => <CentredCaption {...caption} />}
-              style={[
-                styles.leadAssetContainer,
-                isTablet && styles.leadAssetContainerTablet,
-                isTablet && styles.tabletContainer,
-              ]}
+              style={[{ zIndex: 0 }]}
```
c-j-j, line 77: can we object instead of array

c-j-j (approved):
Looks great - nice work

packages/article-magazine-standard/src/article-magazine-standard.js
aronshamash: good catch this some debug code left in. will revert

packages/tooltip/tooltip.tsx
aronshamash: good point. yes will add

packages/article-byline/src/article-byline-with-links.js
aronshamash: `disableTooltip` is prop passed for any templates or component that need to disable it. Within the tooltip component there is a check to only show the tooltip part on tablets.

PR #265, opened 2020-12-02
fix: (TNLT-2130) remove caption under main standard lead image

kamiranoff:
- Remove caption under the lead image in the main standard template

Before:
Screenshot_1606932009
Simulator Screen Shot - iPhone 12 - 2020-12-02 at 18 00 39

After:
Screenshot_1606931305
Screenshot 2020-12-02 at 17 31 01


aronshamash:
@kamiranoff Can you provide an android screenshot ? I take it this requirement is for not just for mobile but tablet too ?

aronshamash:
@kamiranoff btw we normally do a before and after screenshots on our PRs which helps quite often.

Geoff-Ford:
Let's discuss process for pull requests etc in tomorrows meeting.

kamiranoff:
@aronshamash yes this is for `All supported versions of Android and iOS`so tablet and phones. I will add android and before/after screenshot.
I opened the PR a bit early, I also forgot to update the tests. 

I could either leave it as WIP or close it and wait for tomorrows meeting.

packages/article-main-standard/src/article-main-standard.js
```diff
@@ -61,12 +60,6 @@ class ArticlePage extends Component {
                 getImageCrop={getStandardTemplateCrop}
                 onImagePress={onImagePress}
                 onVideoPress={onVideoPress}
-                renderCaption={({ caption }) => (
```
kamiranoff, line 64: Removing this should be fine as there is a default prop for `renderCaption` in `ArticleLeadAsset` but I could also add an extra check within the component to make sure not to call `renderCaption` if the prop is not provided or the prop is provided with something else than a function.

GeorgianSorinMaxim (approved):

c-j-j (approved):

aronshamash:
@kamiranoff can merge this now

aronshamash (approved):

PR #266, opened 2020-12-03
fix: (TNLT-2220) Fix rending of play button on iOS

GeorgianSorinMaxim:
- https://nidigitalsolutions.jira.com/browse/TNLT-2220
- Fix rending of play button on iOS

Before:
Simulator Screen Shot - iPhone 12 - 2020-12-03 at 13 16 23
Simulator Screen Shot - iPhone 12 - 2020-12-03 at 13 16 33

After:
Simulator Screen Shot - iPhone 12 - 2020-12-03 at 13 16 05
Simulator Screen Shot - iPhone 12 - 2020-12-03 at 13 16 09

### DIFF
--- a/packages/article-skeleton/src/body-utils.js
+++ b/packages/article-skeleton/src/body-utils.js
@@ -2,6 +2,15 @@
 import memoize from "memoize-one";
 import { FontStorage } from "@times-components-native/typeset";
 
+export const getAttributes = (attributes) => {
+  if (attributes?.value?.includes("▶")) {
+    attributes.value = attributes.value.replace(/▶/g, "▸");
+    return attributes;
+  }
+
+  return attributes;
+};
+
 export const setupInlineContent = (
   skeletonProps,
   unprocessedContent,
@@ -57,13 +66,17 @@ export const setupInlineContent = (
 
   // Stash the rest of the content to append after
   const afterInlineContent = unprocessedContent.slice(inlineContentEndIndex);
+  const modifiedInlineItemAttributes = getAttributes(inlineItem.attributes);
+  const itemAttributes = modifiedInlineItemAttributes?.attributes
+    ? modifiedInlineItemAttributes.attributes
+    : inlineItem.attributes;
 
   // Add the item and content to inline
   processedContent.push({
     ...inlineItem,
     name: "inlineContent",
     attributes: {
-      ...inlineItem.attributes,
+      ...itemAttributes,
       originalName: inlineItem.name,
       inlineContent,
       skeletonProps,
--- a/packages/article-summary/src/fixContent.js
+++ b/packages/article-summary/src/fixContent.js
@@ -1,0 +1,18 @@
+export default function fix(ast) {
+  if (ast?.length > 0) {
+    const astObj = ast[0];
+    const { children } = astObj;
+    if (children?.length > 0) {
+      children.map((item) => {
+        if (item.attributes?.value?.includes("▶")) {
+          item.attributes.value = item.attributes.value.replace(/▶/g, "▸");
+          return item;
+        }
+        return item;
+      });
+    }
+    return ast;
+  }
+
+  return ast;
+}
--- a/packages/markup-forest/src/markup-forest.js
+++ b/packages/markup-forest/src/markup-forest.js
@@ -6,6 +6,7 @@ export const render = (renderers) => {
     const renderedChildren = children.map((child, index) =>
       run(child, `${key}.${index}`, index),
     );
+
     const result = renderer.call(
       renderers,
       key,
--- a/packages/article-skeleton/src/article-body/article-body-row.js
+++ b/packages/article-skeleton/src/article-body/article-body-row.js
@@ -23,6 +23,7 @@ import styleFactory from "../styles/article-body";
 import ArticleLink from "./article-link";
 import InlineNewsletterPuff from "./inline-newsletter-puff";
 import { useResponsiveContext } from "@times-components-native/responsive";
+import { getAttributes } from "../body-utils";
 
 export default ({
   data,
@@ -68,13 +69,15 @@ export default ({
 
   return {
     text(key, attributes) {
+      const fixedAttributes = getAttributes(attributes);
+
       const attr = {
         tag: "FONT",
         settings: fontConfig.body,
       };
       return new AttributedString(
-        attributes.value,
-        attributes.value.split("").map(() => [attr]),
+        fixedAttributes.value,
+        fixedAttributes.value.split("").map(() => [attr]),
       );
     },
     inline(key, attributes, renderedChildren) {
--- a/packages/article-summary/src/article-summary-content.js
+++ b/packages/article-summary/src/article-summary-content.js
@@ -3,6 +3,7 @@ import { Text } from "react-native";
 import PropTypes from "prop-types";
 import { propTypes as treePropType } from "@times-components-native/markup-forest";
 import { renderAst as defaultRenderAst } from "./article-summary";
+import fixContent from "./fixContent";
 import styles from "./styles";
 
 const ArticleSummaryContent = ({
@@ -29,7 +30,7 @@ const ArticleSummaryContent = ({
       style={[styles.text, style]}
       {...numberOfLinesProp}
     >
-      {renderAst(ast)}
+      {renderAst(fixContent(ast))}
     </Text>
   ) : null;
 };


kamiranoff (approved):

packages/markup-forest/src/markup-forest.js
```diff
@@ -6,10 +17,11 @@ export const render = (renderers) => {
     const renderedChildren = children.map((child, index) =>
       run(child, `${key}.${index}`, index),
     );
+
     const result = renderer.call(
       renderers,
       key,
-      attributes,
+      getAttributes(attributes),
```
c-j-j, line 24: I think this markup-forest module should only be parsing the content's AST and calling the appropriate renderer here (https://github.com/newsuk/times-components-native/blob/master/packages/article-skeleton/src/article-body/article-body-row.js#L69) - I'm not sure if we should have other transformation logic here.

Whenever we've had to transform the article contents, we've done it in this `fixup` call here https://github.com/newsuk/times-components-native/blob/04a14e123c3c6d76253930c563e203c2df1011c0/packages/article-skeleton/src/article-skeleton.js#L85 - I wonder if this logic to convert the play symbols should also be added there.

What do you think?

aronshamash: yes I agree. I think markup-forest is better kept just for parsing the AST

kamiranoff: sorry, but what is the AST?

c-j-j: When we query TPA for the edition data, it returns a tree data structure that we call an AST (abstract syntax tree). This tree gets parsed by the markup-forest module that you've changed, and calls the appropriate renderer. For example with the contents below, when it reads the node with `name` set to `text` - it'll find the `text` renderer in article-body-row.js, which will ultimately render a `<Text>` component with the contents `Ole Gunnar Solskjaer surprised many when`.  

```
                  "summary800": [
                    {
                      "name": "paragraph",
                      "children": [
                        {
                          "name": "text",
                          "attributes": {
                            "value": "Ole Gunnar Solskjaer surprised many when "
                          },
                          "children": []
                        },
                        {
                          "name": "link",
                          "attributes": {
                            "href": "https://www.thetimes.co.uk/article/manchester-united-need-a-new-striker-says-ole-gunnar-solskjaer-jcb3s2wfp",
                            "type": "article",
                            "canonicalId": "manchester-united-need-a-new-striker-says-ole-gunnar-solskjaer-jcb3s2wfp"
                          },
                          "children": [
                            {
                              "name": "text",
                              "attributes": {
                                "value": "he warned on Monday"
                              },
                              "children": []
                            }
                          ]
                        },
```

Does that help?

kamiranoff: yes, very helpful, thanks for the explanation. I think I understand why it is called `markup-forest` now.

packages/markup-forest/src/markup-forest.js
```diff
@@ -1,3 +1,14 @@
+const getAttributes = (attributes) => {
+  if (attributes && attributes.value && attributes.value.includes("▶")) {
```
c-j-j, line 2: FYI - we do have a fairly recent babel config on this project... so you can do `attributes?.value?.includes("xxx")`. Also below you can use object spreading, i.e. `const newAttributes = {...attributes, value: attributes.value.replace('x', 'y')`

packages/markup-forest/src/markup-forest.js
Geoff-Ford: Yep, I agree with @c-j-j here. This should probably go in `body-utils`

GeorgianSorinMaxim: It's not just the article body that has the issue, see the screenshots above. I don't think `article-body-row` would be the only place to fix this.

c-j-j: Good point. Looks like the article summary also has that issue. I think changing packages/article-summary/src/article-summary-content.js will need to be updated too.

I appreciate this now requires a fix in two places instead of one so it's not ideal - maybe we can add a helper function somewhere to remove the duplication.

GeorgianSorinMaxim:
The PR has been updated.

packages/article-summary/src/fixContent.js
```diff
@@ -0,0 +1,18 @@
+export default function fix(ast) {
+  if (ast?.length > 0) {
+    const astObj = ast[0];
+    const { children } = astObj;
+    if (children?.length > 0) {
+      children.map((item) => {
+        if (item.attributes?.value?.includes("▶")) {
```
kamiranoff, line 7: could you reuse your `getAttributes` function here?

packages/article-skeleton/src/body-utils.js
```diff
@@ -2,6 +2,15 @@
 import memoize from "memoize-one";
 import { FontStorage } from "@times-components-native/typeset";
 
+export const getAttributes = (attributes) => {
```
aronshamash, line 5: think this function need to have a name that is more descriptive of what it is actually doing.

aronshamash: also needs some unit tests

aronshamash: maybe `modifyInlineItems` ? `getTransformedInlineItems` ?

packages/article-skeleton/src/body-utils.js
```diff
@@ -57,13 +66,17 @@ export const setupInlineContent = (
 
   // Stash the rest of the content to append after
   const afterInlineContent = unprocessedContent.slice(inlineContentEndIndex);
+  const modifiedInlineItemAttributes = getAttributes(inlineItem.attributes);
+  const itemAttributes = modifiedInlineItemAttributes?.attributes
```
aronshamash, line 70: can't all this logic go in the function above `getAttributes` ?

packages/article-summary/src/fixContent.js
aronshamash: yes definitely need a way to reuse the existing function

packages/article-summary/src/fixContent.js
```diff
@@ -0,0 +1,18 @@
+export default function fix(ast) {
```
aronshamash, line 1: this needs unit tests adding

packages/article-skeleton/src/body-utils.js
c-j-j: Doesn't this need to recursively parse the contents? i.e. if there's bold or italic text, or text that is nested in a keyfact for example - we'd need to drill down to the child and transform the contents.

packages/markup-forest/src/markup-forest.js
GeorgianSorinMaxim: done

packages/article-skeleton/src/body-utils.js
GeorgianSorinMaxim: @c-j-j I haven't seen contents that would need that. If that exists, then it would be good to have examples.

c-j-j: I had to do something similar for the front pages on tablet, where I needed to transform the contents to add hyphens. There's unit tests here: https://github.com/newsuk/times-components-native/blob/61ee55ee22a97b1fb62b456e37e16a1cc206b3b6/packages/utils/__tests__/hyphenate-article-content.test.ts#L16 
and the code is here which does the recursive transform https://github.com/newsuk/times-components-native/blob/61ee55ee22a97b1fb62b456e37e16a1cc206b3b6/packages/utils/src/hyphenate-article-content.ts#L28

I'll try to find an article that has that. Times2 is usually the best section to find such content.

c-j-j: Actually.. this may even not be necessary. Just seen that you're calling this `getAttributes` function by the `text` renderer (https://github.com/newsuk/times-components-native/pull/266/files/0d337d4c2f94133dfc104851ce171e42389ba19c#diff-a66d7b095d0efea33483d62e1bb0690cb7c2b1a086ea4273bca53ef0acfc3a26R72) - which is a child node in the AST - and I think that'll be all that's required to fix the content on the article pages.

I wonder if we could also override the `text` renderer for the article summary to do something similar, which is here https://github.com/newsuk/times-components-native/blob/65701937a376946d064604ff759da3955f0c4560/packages/article-summary/src/article-summary-renderer.js#L6

please let me know if I'm not making sense

packages/article-skeleton/src/body-utils.js
c-j-j: as mentioned here https://github.com/newsuk/times-components-native/pull/266#discussion_r539449704 -  if we fix this issue by just doing the transforms in the `text` renderer - I think this code here can be reverted as the transform is done later on at render-time.

packages/article-summary/src/fixContent.js
c-j-j: this function may not be required if we can just call `getAttributes` from the `text` renderers on the article summary/article pages

GeorgianSorinMaxim: to be honest I'm confused by these suggestions.

c-j-j: Let's jump on a call in the morning? Might be easier to explain

PR #267, opened 2020-12-04
fix: (TNLT-2108): IOS/And: add subscript and superscript for common characters

kamiranoff:
- https://nidigitalsolutions.jira.com/browse/TNLT-2108
- display subscript and superscript characters for characters with the "superscript" and "subscript" attribute name. 

Sorry this is a lot of screenshots, but the changes affect subscript and superscript characters both in ios and android

Before subscript:
Simulator Screen Shot - iPhone 12 - 2020-12-04 at 13 34 58
Before superscript
Screenshot_1607089723

After subscript
Screenshot_1607087537
Simulator Screen Shot - iPhone 12 - 2020-12-04 at 14 13 02

After superscript:
Simulator Screen Shot - iPhone 12 - 2020-12-04 at 13 49 59
Screenshot_1607090248


packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -149,20 +150,12 @@ export default ({
     },
     subscript(key, attributes, children) {
       const childStr = AttributedString.join(children);
-      const attr = {
-        tag: "FONT",
-        settings: fontConfig.body,
-      };
-      childStr.addAttribute(0, childStr.length, attr);
+      childStr.string = toSubscript(childStr.string);
```
kamiranoff, line 153: I am not sure this is the best place to do this or the best way of doing it, let me if this can be improved.

packages/utils/__tests__/subscript-superscript-converter.test.ts
```diff
@@ -0,0 +1,59 @@
+import { subscripts, superscripts, toSubscript, toSuperscript } from "../";
+
+describe("should subscript characters", () => {
```
kamiranoff, line 3: Also if you have more meaningful tests in mind, I could add more.

packages/utils/src/subscript-superscript-mapper.ts
```diff
@@ -0,0 +1,69 @@
+export const superscripts = {
```
kamiranoff, line 1: This solution might be more limited but I thought it is nicer than playing with font size and line height

packages/utils/src/subscript-superscript-mapper.ts
```diff
@@ -0,0 +1,69 @@
...
+  "7": "⁷",
+  "8": "⁸",
+  "9": "⁹",
+  "+": "⁺",
+  "-": "⁻",
+  "=": "⁼",
+  "(": "⁽",
+  ")": "⁾",
+  n: "ⁿ",
+  i: "ⁱ",
```
c-j-j, line 18: Just out of interest, what's the reason for not having all letters here? I don't fully understand the context, but I would have assumed that every letter in `subscript`/`superscript` should be smaller.

packages/utils/src/subscript-superscript-mapper.ts
c-j-j: Agreed. This seems like a good solution to me.

packages/utils/src/subscript-superscript-mapper.ts
kamiranoff: The issue is that the subscript/superscript characters are not rendering in mobile:
The web version using `sup` and `sub` tags as the AST returns the characters in section with `name` of `subscript` and `superscript`. And `sup` and `sub` would work for most characters I believe.

So I took characters from the `Unicode block`:
https://en.wikipedia.org/wiki/Superscripts_and_Subscripts_(Unicode_block)

But looking at this table I should also add the characters from this section:
https://en.wikipedia.org/wiki/Unicode_subscripts_and_superscripts#Latin_and_Greek_tables

Here are a couple of example articles with superscript and subscript characters:
https://www.thetimes.co.uk/article/how-babylonians-beat-pythagoras-by-1-000-years-wjlfp6rqm
https://www.thetimes.co.uk/article/scientists-confirm-accuracy-of-einsteins-theory-vk6fkz27bwq

c-j-j: Ah I see - it makes sense to only have a subset of the chars as you've done

aronshamash (approved):
looks good!

c-j-j (approved):

PR #269, opened 2020-12-07
chore: Fix max tablet width

Geoff-Ford:
Seems we have the max tablet width currently set to `1194`. As far as I can tell this is because this is the landscape width for iPad Pro 11. However the iPad Pro 12.9 is `1366`  thus the PR

Before:
Screenshot 2020-12-07 at 11 43 37

After:
Screenshot 2020-12-07 at 11 29 09

### DIFF
--- a/packages/article-in-depth/__tests__/android/__snapshots__/article-tablet.android.test.js.snap
+++ b/packages/article-in-depth/__tests__/android/__snapshots__/article-tablet.android.test.js.snap
@@ -8,7 +8,7 @@ exports[`1. article in depth - tablet 1`] = `
         "alignItems": "center",
         "alignSelf": "center",
         "backgroundColor": "rgba(60, 129, 190, 1)",
-        "maxWidth": 1194,
+        "maxWidth": 1366,
         "paddingBottom": 40,
         "paddingHorizontal": 20,
         "paddingTop": 45,
@@ -69,7 +69,7 @@ exports[`1. article in depth - tablet 1`] = `
       Object {
         "alignSelf": "center",
         "marginBottom": 0,
-        "maxWidth": 1194,
+        "maxWidth": 1366,
         "width": "100%",
       }
     }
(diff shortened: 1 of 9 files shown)


packages/styleguide/src/breakpoints/index.shared.js
```diff
@@ -32,7 +32,7 @@ export default {
   huge: 1320,
   medium: 768,
   nativeTablet: 660,
-  nativeTabletWide: 1194,
+  nativeTabletWide: 1366,
```
aronshamash, line 35: think this should be `maxTabletWidth` ?

aronshamash: think having word `native` is odd now we have no web specific code.

packages/article-in-depth/__tests__/android/__snapshots__/article-tablet.android.test.js.snap
```diff
@@ -8,7 +8,7 @@ exports[`1. article in depth - tablet 1`] = `
         "alignItems": "center",
         "alignSelf": "center",
         "backgroundColor": "rgba(60, 129, 190, 1)",
-        "maxWidth": 1194,
+        "maxWidth": 1366,
```
c-j-j, line 11: This looks like it'll do the same as setting `width: "100%` - might be worth trying that in case apple release an even bigger iPad?

packages/styleguide/src/breakpoints/index.shared.js
Geoff-Ford: The same could be said for `nativeTablet` above it. But this is a fix to get in to todays release so I don't want to start refactoring.

packages/article-in-depth/__tests__/android/__snapshots__/article-tablet.android.test.js.snap
c-j-j: or can leave it as is to get a fix out to beta users, and look into this more another time

Geoff-Ford: We could be we have always been explicit. 🤷  I'm easy to be honest. `100%` should also work.

aronshamash (approved):

c-j-j (approved):

kamiranoff (approved):

PR #273, opened 2020-12-08
TNLT-995-add-image-placeholders

GeorgianSorinMaxim:
- https://nidigitalsolutions.jira.com/browse/TNLT-995
- Fix rendering image placeholders for offline use on iOS


GeorgianSorinMaxim:
Failing test
```Summary of all failing tests
 FAIL  packages/responsive-image/__tests__/image.test.tsx
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /home/circleci/project/node_modules/react-native-image-zoom-viewer/built/image-viewer.component.js:415
                    return <react_native_1.View key={index} style={{ width: screenWidth, height: screenHeight }}/>;
                           ^

    SyntaxError: Unexpected token <

      at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:537:17)
      at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:579:25)
      at Object.<anonymous> (node_modules/react-native-image-zoom-viewer/src/index.ts:1:1)

**Comment by c-j-j** [comment]
Hi @GeorgianSorinMaxim - if you add `jest.mock("react-native-image-zoom-viewer", () => "ImageZoomView");` to the test `packages/responsive-image/__tests__/image.test.tsx`  that should fix

File: packages/article-image/src/article-image.base.js [hunk 1]
```diff
@@ -73,6 +74,10 @@ const ArticleImage = ({
   const { caption, credits } = captionOptions;
 
   if (display === "inline") {
+    if (uri === "") {
```
kamiranoff, line 77: I guess here too if uri is undefined/null you want to return the `Placeholder`

packages/author-profile/src/author-profile-head-image.js
```diff
@@ -1,17 +1,22 @@
...
-  <Image
-    aspectRatio={1}
-    style={styles.authorPhoto}
-    uri={uri}
-    rounded
-    accessibilityLabel="author-image"
-  />
-);
+const AuthorProfileHeadImage = ({ uri }) => {
+  return uri && uri !== "" ? (
```
kamiranoff, line 8: here too `return uri ?` should suffice

aronshamash:
@GeorgianSorinMaxim Can you please add some screen shots to the desc

packages/article-comment-tablet/src/article-left-column/article-left-column.js
```diff
@@ -21,7 +22,11 @@ const ArticleLeftColumn = ({
 }) => (
   <View style={styles.leftColumnContainer}>
     <View style={styles.authorContainer}>
-      <Image aspectRatio={1} uri={authorImage} rounded />
+      {authorImage === "" ? (
```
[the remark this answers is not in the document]
aronshamash: or swap the logic so ternary is `authorImage ? Image : Placeholder`

packages/image/src/styles/shared.js
```diff
@@ -81,6 +81,8 @@ const styles = {
     width: "100%",
   },
   placeholder: {
+    zIndex: 999,
+    elevation: 5,
```
aronshamash, line 85: Is elevation used by ios ? I thought only Android ?

packages/responsive-image/src/index.tsx
```diff
@@ -129,18 +130,36 @@ const ResponsiveImage = (props: ResponsiveImageProps) => {
...
-          setShowOnline(true);
-          setShowOffline(false);
-          setShowPlaceholder(false);
-          return;
-        }
-        if (offlineUrl in results) {
+      cache
+        .then((results) => {
+          if (onlineUrl in results) {
+            setShowOnline(true);
```
aronshamash, line 136: think some of this could be DRYed up a bit maybe a function which takes 3 param and does 
          
  `setShowOnline();setShowOffline(); setShowPlaceholder();`
?

packages/author-profile/src/author-profile-head-image.js
GeorgianSorinMaxim: maybe `return uri && uri === ""`

packages/article-image/src/article-image.base.js
GeorgianSorinMaxim: sure

packages/article-comment-tablet/src/article-left-column/article-left-column.js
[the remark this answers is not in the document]
GeorgianSorinMaxim: not seeing benefits in these simple changes, but why not

packages/image/src/styles/shared.js
GeorgianSorinMaxim: used by Android but I assume this is used by both projects

packages/responsive-image/src/index.tsx
GeorgianSorinMaxim: I think this is out of the scope of this bug fix

packages/author-profile/src/author-profile-head-image.js
kamiranoff: yeah, was just pointing out that empty string is also a falsy value.

GeorgianSorinMaxim:
PR is updated

packages/author-profile/src/author-profile-head-image.js
GeorgianSorinMaxim: true, that's changed now

packages/responsive-image/src/index.tsx
c-j-j: I don't think you'd be able to move the `setShowOnline` etc into a function @aronshamash  as they're from a hook. 

I do agree that there's a need to refactor some of this out as there's a lot of duplication.  I think`useReducer` would be the best way to fix this, you could dispatch actions to change all three of these state variables. It'll be easier to unit test too.

GeorgianSorinMaxim:
PR is now updated to address comments

packages/responsive-image/src/index.tsx
aronshamash: I think we should try and improve it as long as it doesn't take too long to do.

GeorgianSorinMaxim: after more testing, this fix is not fit for all use cases. the original code is flaky and the logic is wrong and confusing to say the least.

GeorgianSorinMaxim: the PR was updated

packages/image/src/styles/shared.js
aronshamash: just that the title had (ios) in so I thought the whole thing was for ios only

packages/article-comment-tablet/src/article-left-column/article-left-column.js
```diff
@@ -21,7 +22,11 @@ const ArticleLeftColumn = ({
 }) => (
   <View style={styles.leftColumnContainer}>
     <View style={styles.authorContainer}>
-      <Image aspectRatio={1} uri={authorImage} rounded />
+      {authorImage !== "" ? (
```
c-j-j, line 25: Could an authorImage ever be `null`/`undefined`? I wonder if it's safer to check if it's falsey instead with `authorImage ? (etc)`

c-j-j (approved):

kamiranoff (approved):

aronshamash:
@GeorgianSorinMaxim need to update snaps in the `icons` package

PR #282, opened 2020-12-17
fix: (TNLT-3070) Fix crashes for unloaded cells

gsm-dev-x-ldn:
https://nidigitalsolutions.jira.com/browse/TNLT-3070
Fix crashes for unloaded cells
- There are no visual changes, so I didn't include screenshots

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
     "react-native-device-info": "5.3.1",
     "react-native-image-zoom-viewer": "2.2.26",
     "react-native-navigation": "^6.11.0",
-    "react-native-svg": "9.11.1",
+    "react-native-svg": "12.1.0",
     "react-native-v8": "0.63.2-patch.1",
     "react-native-webview": "10.9.0",
     "unfetch": "^4.1.0",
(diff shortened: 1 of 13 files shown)


kamiranoff (approved):

package.json
```diff
@@ -96,7 +96,7 @@
     "react-native-device-info": "5.3.1",
     "react-native-image-zoom-viewer": "2.2.26",
     "react-native-navigation": "^6.11.0",
-    "react-native-svg": "9.11.1",
+    "react-native-svg": "12.1.0",
```
c-j-j, line 99: You'll need to also update this here https://github.com/newsuk/nuk-tnl-app-ios-universal/blob/develop/package.json#L15 on both their `develop` and `epic/contemporary-ua` branches. Would be best to liaise with Dan Bennett/Lee Daffen on best approach for that.

gsm-dev-x-ldn: This is what Lee said: `Probably the best thing to do is publish the new version of TCN and include the required dependency update in the release notes. We can then make the required change to the package manifest the next time we update the version of TCN in the app.`

PR #284, opened 2020-12-18
fix: (TNLT-5493) Update dependencies for Xcode 12 - fix crash in icons

kamiranoff:
- https://nidigitalsolutions.jira.com/browse/TNLT-5493
- update dependencies for Xcode 12
- Fix crashes in icons.

I managed to make the ios app compile with Xcode 12 without removing `@react-native-community/art`. So I made a simplified [PR in IOS](https://github.com/newsuk/nuk-tnl-app-ios-universal/pull/2242).

In this PR I only migrated 3 icons from `@react-native-community/art` to `react-native-svg` (only the one that caused a crash).


packages/article-skeleton/__tests__/android/__snapshots__/inline-newsletter-puff.test.js.snap
```diff
@@ -1403,132 +1161,17 @@ exports[`Inline Newsletter Puff renders the success view after subscribing:  wid
...
-                      0,
-                      1,
-                      0,
-                      0,
-                    ]
-                  }
-                  opacity={1}
-                  originX={0}
-                  originY={0}
+                  fill={-16750951}
```
c-j-j, line 1166: We should check if the inline newsletter puff still works on android now that we've updated this library. We've previously seen the app crashing at runtime after updating these SVG libs.

kamiranoff: Happy new year :) 
do you know how I could test this?

c-j-j: Happy new year!

if you navigate to this article https://www.thetimes.co.uk/edition/news/wait-40-years-for-another-scottish-independence-vote-says-boris-johnson-kwb7njq99 - you'll see the red box puff after several paragraphs... it would be good to check it renders okay, and pressing subscribe to check it renders the confirmation too (FYI you can unsubscribe here https://home.thetimes.co.uk/myNews)

kamiranoff: Good catch @c-j-j! 
I have a crash on android. Not in the times component native where icons render fine, but in the native android app. I am investigating. 

Seems to be related to this issue: https://github.com/react-native-svg/react-native-svg/issues/1405

kamiranoff: @c-j-j it might be because android is built using the current version of `times-component-native` which doesn't have the updated version of `react-native-svg` Can I build android locally with this branch of `times-components-native`?

c-j-j: @kamiranoff  You can - the easiest way I can think of is to create a beta release from this branch. If you update the package.json version to something like `2.0.3-beta` - and run the circle CI task `run_tests/hold_release_beta` to deploy a beta version to JFrog.

You will will need to change this line to use `https://newsint.jfrog.io/newsint/times-components-android-artifacts-beta` :

https://github.com/newsuk/nuk-tnl-app-android-universal/blob/6b67580221e240800a4c241caab490d7fa8d5a9a/build.gradle#L50 

And change this line to pull the version `2.0.3-beta`:

https://github.com/newsuk/nuk-tnl-app-android-universal/blob/7f66b822f6c2a5aa59365c5fa89809ee7454522e/dependencies.gradle#L32

That hopefully should be everything you need. 

I'm not sure if this process is documented anywhere, but it probably should be!

kamiranoff: Awesome thanks. I ll give it a try

kamiranoff: Yep. Seems to have fixed the issue and it seems to me that Red Box is rendering correctly
Screenshot_1609774244
Screenshot_1609774255

I guess I can revert the version change now

c-j-j: Yes exactly - you can revert that version back. 

Good to see that is still working though 👍

c-j-j (approved):

gsm-dev-x-ldn (approved):

PR #285, opened 2021-01-05
chore: Remove ad variants

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-6278

* Removed logic in variant testing related specifically to the article and section ads.
* Set default article ads to always use single MPU ads (this will become dynamic in the next ticket https://nidigitalsolutions.jira.com/browse/TNLT-5829).
* Set default section ads to always use leaderboards on portrait and billboards on landscape.
* Changed slotnames to something more meaningful (`native-single-mpu`, `native-double-mpu`, `native-section-ad`)
* Fixed tests


packages/article-skeleton/src/body-utils.js
```diff
@@ -82,14 +83,14 @@ export const setupInlineContent = (
...
-  articleMpu,
+const setupArticleMpuAd = (
   currentAdSlotIndex,
   contentWithoutAdSlot,
   skeletonProps,
 ) => {
-  const { adPosition, group, width, height, slotName } = articleMpu;
-  const isControlGroup = group === "A";
+  const adPosition = 5;
+  const slotName = "native-single-mpu";
```
Geoff-Ford, line 92: The above are to be decided dynamically in the next ticket https://nidigitalsolutions.jira.com/browse/TNLT-5829 so this is temporary (used to come from the variant).

packages/ad/src/utils/ad-init.js
```diff
@@ -3,7 +3,7 @@
 // NOTE: this function must be self-contained, i.e. contain no references to variables
 // defined outside the function, so that it can be passed into a WebView.
 export default ({ el, data, platform, eventCallback, window }) => {
-  const nativeAdRegex = /^native-(inline|section)-/;
+  const nativeAdRegex = /^native-(inline|section|(single|double)-mpu)/;
```
aronshamash, line 6: are we going to support both single and double mpu then ?

Geoff-Ford: Yep, dynamically based on the article content.

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -82,10 +80,10 @@ const ArticleWithContent = (props) => {
     [],
   );
 
-  const fixedContent = useMemo(
-    () => [...fixup(props, variants), { name: "footer" }],
-    [content, isTablet],
-  );
+  const fixedContent = useMemo(() => [...fixup(props), { name: "footer" }], [
```
c-j-j, line 83: It's a shame we're losing all this logic since we don't have any current A/B tests. I guess we can always refer back to this PR if we ever need to add it back in

packages/article-skeleton/src/body-utils.js
```diff
@@ -82,14 +83,14 @@ export const setupInlineContent = (
...
-const setupArticleMpuTestAd = (
-  articleMpu,
+const setupArticleMpuAd = (
   currentAdSlotIndex,
   contentWithoutAdSlot,
   skeletonProps,
 ) => {
-  const { adPosition, group, width, height, slotName } = articleMpu;
-  const isControlGroup = group === "A";
+  const adPosition = 5;
```
c-j-j, line 91: Just a thought... now that we're not doing A/B testing with the ad positioning, I wonder if we should revert back to using the position that TPA provides rather than hard coding `5` in the front end. They currently return an `ad` node in the 5th position, which we could use. Just thinking if they ever change the ad position on the backend, this won't get reflected here.

c-j-j (approved):

packages/ad/src/utils/ad-init.js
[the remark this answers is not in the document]
Geoff-Ford: Nope, that is the original format which is used by mobile

packages/article-skeleton/src/article-skeleton.js
Geoff-Ford: That was my thinking also. There isn't really a legitimate way of keeping this code around as specific to those tests.

packages/article-skeleton/src/body-utils.js
Geoff-Ford: See my comment on that code about the next ticket.

PR #287, opened 2021-01-06
feat: Dynamic article MPU selection

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-5829

Made article mpu type dependent on if non-paragraph content found within the first 10 items of an article.
* Single MPU if non-paragraph content found within the threshold
* Double MPU if not

https://user-images.githubusercontent.com/5103512/103775064-71e86600-5025-11eb-8fd3-c78eb498d6fb.mov

### DIFF
--- a/packages/article-skeleton/src/body-utils.js
+++ b/packages/article-skeleton/src/body-utils.js
@@ -89,14 +89,19 @@ const setupArticleMpuAd = (
   skeletonProps,
 ) => {
   const adPosition = 5;
-  const slotName = "native-single-mpu";
-  const { height, width } = sizeMap[slotName][0];
+  const doubleMPUThreshold = 10;
 
   // Get index of nth (adPosition) paragraph
   let nthParagraphIndex = currentAdSlotIndex;
 
+  let hasNonParagraphContentBeforeThreshold = false;
+
   contentWithoutAdSlot.reduce((count, item, index) => {
-    if (item.name !== "paragraph") return count;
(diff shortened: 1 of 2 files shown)


packages/article-skeleton/src/body-utils.js
```diff
@@ -89,14 +89,19 @@ const setupArticleMpuAd = (
   skeletonProps,
 ) => {
   const adPosition = 5;
```
Geoff-Ford, line 91: @c-j-j  Re your comment yesterday on `adPosition` and using the default position provided by TPA. Forgot to mention that we now ensure the ad is displayed after `adPosition` number of actual paragraphs regardless of any other content like images and interactives that might be present in

c-j-j: Ah right, fair enough. So it's not a good idea to infer that position from the TPA content as they may not have factored in the images/etc.

aronshamash: ok so are we saying we going to ignore ad position TPA sends us ?

Geoff-Ford: We have been since the introduction of inline ads. Mobile however doesn't get touched so uses the default TPA position.

aronshamash (approved):

kamiranoff (approved):

c-j-j (approved):

PR #288, opened 2021-01-07
fix: (TNLT-995) [Revert] Add image placeholders for offline use (iOS)

kamiranoff:
Reverts newsuk/times-components-native#273

I believe this causing some images not to display
Simulator Screen Shot - iPhone 12 Pro - 2021-01-07 at 09 59 43


c-j-j (approved):

Geoff-Ford:
Have you checked in with @GeorgianSorinMaxim ? He was looking into a fix for this. If it is something minor then easier to add a small fix than revert the whole PR. Unless it is urgent?

aronshamash (approved):

kamiranoff:
> Have you checked in with @GeorgianSorinMaxim ? He was looking into a fix for this. If it is something minor then easier to add a small fix than revert the whole PR. Unless it is urgent? ie. has this made it out into the wild?

I think he is ok with. We could wait for a fix, but I think Shariff wants a build to test. So up to @GeorgianSorinMaxim ?

PR #289, opened 2021-01-07
fix: (TNLT-994) AND/IOS Update play button to new design

kamiranoff:
- Change play video icon to the new design. (https://www.figma.com/file/iLZzyu9pI42Ch8AkIVhfhA/TNLT-994%E2%80%93Play-Icon?node-id=1%3A6)

before:
Simulator Screen Shot - iPhone 12 Pro - 2021-01-07 at 12 17 37
Screenshot_1610021882

after:
Screenshot_1610019994
Screenshot_1610020000


packages/article-lead-asset/article-lead-asset.showcase.js
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { Text } from "react-native";
+import { Dimensions, Text } from "react-native";
```
aronshamash, line 2: There is a `getDimensions` from the `utils` package that you could use btw

packages/video/src/play-icon.native.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { FC } from "react";
+import Svg, { Path, Rect } from "react-native-svg";
+
+export enum PLAY_ICON_SIZE {
```
aronshamash, line 4: I like this style of setting constants. Should we move in this direction @Geoff-Ford @c-j-j ? We have a range of ways but this is my preferred but hadn't done because they tend to be done using lower camel case consts.

packages/video/src/play-icon.native.tsx
```diff
@@ -0,0 +1,38 @@
...
+  MEDIUM = 48,
+  LARGE = 64,
+  XLARGE = 76,
+}
+
+export type PlayIconProps = {
+  containerWidth?: number;
+};
+
+const getSize = (containerWidth: number) => {
```
aronshamash, line 15: could use an object literal here to remove logic?

packages/video/src/play-icon.native.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { FC } from "react";
```
aronshamash, line 1: we don't need a `.native` file as this repo is native only.

packages/video/src/play-icon.native.tsx
```diff
@@ -0,0 +1,38 @@
...
+    case containerWidth > 563:
+      return PLAY_ICON_SIZE.LARGE;
+    case containerWidth > 755:
+      return PLAY_ICON_SIZE.XLARGE;
+    default:
+      return PLAY_ICON_SIZE.SMALL;
+  }
+};
+
+const PlayIcon: FC<PlayIconProps> = ({ containerWidth = 0 }) => {
```
aronshamash, line 28: is this how we want to go forward with icons in situ rather than in `icons` ?

aronshamash:
@kamiranoff looks much better!

packages/video/src/play-icon.native.tsx
kamiranoff: This was already in the video folder. it shows as a new file because I changed from .js to .tsx. 
I think it makes sense to have it within the video folder, but it also makes sense to put it in icons.

packages/video/src/play-icon.native.tsx
c-j-j: Yes, happy with this convention. It makes it clear that this is a global constant whenever it's used.

packages/video/src/play-icon.native.tsx
kamiranoff: I am not sure I understand how?

packages/video/src/play-icon.native.tsx
kamiranoff: I realised this doesn't need to be exported anymore btw.

packages/video/src/play-icon.native.tsx
kamiranoff: I could also put this function inside the component to avoid passing a parameter. Wasn't sure which approach was best

packages/video/src/play-icon.native.tsx
Geoff-Ford: Yep, best for "global" constants. Not necessary for local ones.

packages/video/src/play-icon.tsx
```diff
@@ -0,0 +1,38 @@
...
+    case containerWidth > 270:
+      return PLAY_ICON_SIZE.MEDIUM;
+    case containerWidth > 563:
+      return PLAY_ICON_SIZE.LARGE;
+    case containerWidth > 755:
+      return PLAY_ICON_SIZE.XLARGE;
+    default:
+      return PLAY_ICON_SIZE.SMALL;
+  }
+};
```
Geoff-Ford, line 26: Won't everything above 270 get `PLAY_ICON_SIZE.MEDIUM`? Think the order needs to be reversed i.e. start with largest getting smaller?

Geoff-Ford: Guess this also shows we need some tests for this also to capture all the variations.

packages/video/src/play-icon.native.tsx
aronshamash: I think it depends on if it is reusable. Going forward I prefer they are close to where they are used so keep here. If any `icons` used in multiple place then `icons` makes more sense

packages/video/src/play-icon.tsx
kamiranoff: oh.. yes, good point, that is not gonna work. I ll change this.

packages/video/src/play-icon.native.tsx
aronshamash: actually I think it is fine outside with para. Forget about putting into object literal don't think it will make it easier to read. Just needs a test maybe ?

packages/video/src/play-icon.tsx
kamiranoff: I have fixed the logic.

aronshamash (approved):
looks good

c-j-j (approved):
Looks good!

Geoff-Ford (approved):

PR #291, opened 2021-01-07
feat: (TNLT-3187) Article read indicator

aronshamash:
- Article Read UI treatment only
- https://nidigitalsolutions.jira.com/browse/TNLT-3187
- Native will send over an array of `readArticles` to the `section` page which is put in context.
- Most of the changes are in `packages/edition-slices/src/tiles/shared/tile-summary.tsx` 
- Article Flags don't use render props like the rest of the components but a refactor to make consistent with rest would be fairly large and touch lots of code so hence the need to do the flags animation in `ArticleSummary` not `TileSummary`
- Tablet only feature
- Only edition slices
-  Delay of `500` ms before animation
- Animation duration of `500` ms

### Interaction
https://user-images.githubusercontent.com/1736782/104214423-528b7780-542f-11eb-9651-3ec324ad61ac.mov

### iOS 
Screenshot 2021-01-11 at 16 31 10
Screenshot 2021-01-11 at 16 27 03

### Android 
Screenshot 2021-01-08 at 11 20 29


packages/edition-slices/src/tiles/shared/tile-summary.tsx
```diff
@@ -0,0 +1,197 @@
...
+  ArticleSummaryStrapline,
+} from "@times-components-native/article-summary";
+import { SectionContext } from "@times-components-native/context";
+import { ArticleFlags } from "@times-components-native/article-flag";
+import { colours } from "@times-components-native/styleguide";
+import { ResponsiveContext } from "@times-components-native/responsive";
+import PositionedTileStar from "./positioned-tile-star";
+
+interface Props {
+  bylines?: any;
```
aronshamash, line 15: TODO is all this typing

packages/pages/src/section/section.js
```diff
@@ -142,6 +142,19 @@ class SectionPage extends Component {
       sectionName: section.name,
     });
 
+    const readArticles = [
+      "703d9a72-4f86-11eb-ad71-ea6bb4a570af",
```
aronshamash, line 146: These will come from Native layer but that will be in another future integration ticket. I will remove these before it is merged. This is just for testing.

packages/article-summary/src/article-summary.js
```diff
@@ -103,7 +124,9 @@ ArticleSummary.propTypes = {
     isVideo: PropTypes.bool,
     title: PropTypes.string,
     hide: PropTypes.bool,
+    markAsRead: PropTypes.bool,
   }),
+  saveStar: PropTypes.node,
```
aronshamash, line 129: some other missing `PropTypes`

packages/article-summary/src/article-summary.js
```diff
@@ -21,17 +21,38 @@ function renderAst(ast) {
 }
 
 function ArticleSummaryLabel(props) {
-  const { hide, title, isVideo } = props;
+  const { markAsRead, hide, title, isVideo } = props;
+  const [labelOpacity] = useState(new Animated.Value(1));
```
kamiranoff, line 25: I think you could also use `useRef` here if you don't need `setLabelOpacity` anywhere:
https://reactnative.dev/docs/animated#example
But I am not sure if it will change anything.

packages/edition-slices/src/tiles/shared/index.js
```diff
@@ -1,7 +1,7 @@
 import { getTileImage, getTileSummary, getTileStrapline } from "./utils";
 import TileLink from "./tile-link";
 import TileStar from "./tile-star";
-import TileSummary from "./tile-summary";
+import TileSummary from "./tile-summary.tsx";
```
kamiranoff, line 4: do you need the extension here?

packages/edition-slices/src/tiles/shared/tile-summary.tsx
```diff
@@ -0,0 +1,202 @@
...
+          {({ readArticles }) => {
+            const markAsRead = shouldMarkAsRead(isTablet, readArticles, id);
+
+            if (markAsRead) {
+              Animated.timing(textOpacity, {
+                delay: readArticleAnimationDelay,
+                duration: readArticleAnimationDuration,
+                toValue: 0.6,
+                useNativeDriver: false,
+              }).start();
```
kamiranoff, line 171: should this be in a `useEffect` instead to be triggered only if `markAsRead` changes to true?

packages/edition-slices/src/tiles/shared/tile-summary.tsx
```diff
@@ -0,0 +1,202 @@
...
+import { SectionContext } from "@times-components-native/context";
+import { ArticleFlags } from "@times-components-native/article-flag";
+import { colours } from "@times-components-native/styleguide";
+import { ResponsiveContext } from "@times-components-native/responsive";
+import PositionedTileStar from "./positioned-tile-star";
+
+interface Props {
+  bylines: BylineInput[];
+  bylineStyle?: StyleProp<ViewStyle>;
+  flagColour?: any;
```
kamiranoff, line 22: I guess object would be a little better

packages/edition-slices/src/tiles/shared/tile-summary.tsx
```diff
@@ -0,0 +1,202 @@
...
+  flagsStyle?: StyleProp<ViewStyle>;
+  headlineStyle?: StyleProp<ViewStyle>;
+  labelColour?: string;
+  linesOfTeaserToRender?: number;
+  strapline?: string;
+  straplineStyle?: StyleProp<ViewStyle>;
+  style?: StyleProp<ViewStyle>;
+  summary?: Markup;
+  summaryStyle?: any;
```
kamiranoff, line 31: I think this could be `StyleProp<TextStyle>`

packages/article-summary/src/article-summary.js
```diff
@@ -21,17 +21,38 @@ function renderAst(ast) {
...
   if (hide || (!title && !isVideo)) {
     return null;
   }
 
-  return (
+  if (markAsRead) {
+    Animated.timing(labelOpacity, {
+      duration: 500,
+      toValue: 0.6,
+      useNativeDriver: false,
```
c-j-j, line 35: what's the reason for setting this to `false`? Having a quick look at their docs, as we're animating an `opacity`value  it seems that we could set this to `true`...

packages/edition-slices/src/tiles/shared/tile-summary.tsx
```diff
@@ -0,0 +1,202 @@
...
+  const readArticleAnimationDuration = 300;
+  const readArticleAnimationDelay = 500;
+
+  type MarkAsReadProps = {
+    children: ReactNode;
+    markAsRead: boolean;
+  };
+
+  const MarkAsRead = ({ children, markAsRead }: MarkAsReadProps) => (
+    <>
```
c-j-j, line 88: is this fragment wrapper needed?

packages/edition-slices/src/tiles/shared/tile-summary.tsx
```diff
@@ -0,0 +1,202 @@
...
+      <MarkAsRead markAsRead={markAsRead}>
+        <ArticleSummaryStrapline strapline={strapline} style={straplineStyle} />
+      </MarkAsRead>
+    );
+
+  const shouldMarkAsRead = (
+    isTablet: boolean,
+    readArticles: Array<string> | null,
+    articleId: string,
+  ) => !!(isTablet && readArticles && readArticles.includes(articleId));
```
c-j-j, line 156: Did you notice any performance delay on the News section with this? If every tile is doing this on a large section, it could add a delay to the user - something to keep an eye on - and to add a `useMemo` if required.

packages/edition-slices/src/tiles/shared/tile-summary.tsx
```diff
@@ -0,0 +1,202 @@
...
+      {({ isTablet }) => (
+        <SectionContext.Consumer>
+          {({ readArticles }) => {
+            const markAsRead = shouldMarkAsRead(isTablet, readArticles, id);
+
+            if (markAsRead) {
+              Animated.timing(textOpacity, {
+                delay: readArticleAnimationDelay,
+                duration: readArticleAnimationDuration,
+                toValue: 0.6,
```
c-j-j, line 169: We have this value ( and the 500ms transition duration) duplicated in two places... is it worth extracting somewhere globally?

packages/edition-slices/src/tiles/shared/tile-summary.tsx
```diff
@@ -0,0 +1,202 @@
...
+              />
+            );
+          }}
+        </SectionContext.Consumer>
+      )}
+    </ResponsiveContext.Consumer>
+  );
+};
+
+export default TileSummary;
```
c-j-j, line 202: nice work translating this to Typescript

c-j-j (approved):
Couple of minor comments but overall the approach looks fine

packages/edition-slices/src/tiles/shared/index.js
aronshamash: no I think IDE added it helpfully !

packages/edition-slices/src/tiles/shared/tile-summary.tsx
aronshamash: for TS yes

packages/edition-slices/src/tiles/shared/tile-summary.tsx
aronshamash: no perf delay noticed but will keep an eye on it.

packages/article-summary/src/article-summary.js
aronshamash: Ah good point thanks I'll change. Shouldn't change anything.

packages/edition-slices/src/tiles/shared/tile-summary.tsx
aronshamash: yes I thought of this but couldn't think of the best place. Maybe just exported from `TileSummary` ? Any suggestions @c-j-j ?

c-j-j: Potentially the styleguide?

packages/edition-slices/src/tiles/shared/tile-summary.tsx
aronshamash: Can't use `Object` as a type. Nearest thing is `Record<string, unknown>`

packages/edition-slices/src/tiles/shared/tile-summary.tsx
aronshamash: unfortunately not because there is a `summaryStyle.lineHeight` which doesn't exist in that type

packages/edition-slices/src/tiles/shared/tile-summary.tsx
kamiranoff: ah, I think it's even better. At least it makes clear it is not a string that is expected.

packages/article-summary/src/article-summary.js
aronshamash: yes set to `true` now

packages/edition-slices/src/tiles/shared/tile-summary.tsx
aronshamash: Even that is an issue as `Property 'color' is missing in type`. this was the reason I left as `any` :(

kamiranoff: and what about:
```
{
 color: string
}
```

wouldn't that work?

Geoff-Ford:
Can you add any screenshots?

packages/edition-slices/src/tiles/shared/tile-summary.tsx
aronshamash: OK I've added to styleguide.

packages/edition-slices/src/tiles/shared/tile-summary.tsx
aronshamash: no see `typecheck` output:

```
yarn run v1.16.0
$ tsc --project ./
packages/edition-slices/src/tiles/shared/tile-summary.tsx:152:8 - error TS2322: Type '{ style: StyleProp<ViewStyle>; flags: (ExpirableFlag | null)[] | null | undefined; longRead: boolean | null | undefined; color: string; } | { ...; }' is not assignable to type 'IntrinsicAttributes & Pick<Pick<{ flags: any; longRead: any; color: any; style: any; withContainer: any; }, never> & Pick<InferProps<{ flags: Requireable<(InferProps<{ expiryTime: Requireable<string>; type: Requireable<...>; }> | null | undefined)[]>; longRead: Requireable<...>; withContainer: Requireable<...>; }>, ...'.
  Property 'color' is missing in type '{ style: StyleProp<ViewStyle>; flags: (ExpirableFlag | null)[] | null | undefined; longRead: boolean | null | undefined; }' but required in type 'Pick<Pick<{ flags: any; longRead: any; color: any; style: any; withContainer: any; }, never> & Pick<InferProps<{ flags: Requireable<(InferProps<{ expiryTime: Requireable<string>; type: Requireable<string>; }> | null | undefined)[]>; longRead: Requireable<...>; withContainer: Requireable<...>; }>, "longRead" | ... 1 ...'.

152       <ArticleFlags
           ~~~~~~~~~~~~


Found 1 error.
```

aronshamash:
> Can you add any screenshots?

done

kamiranoff (approved):

packages/edition-slices/src/tiles/shared/tile-summary.tsx
kamiranoff: Ah yes, it is strange. TypeScript thinks `color` is a required argument of `ArticleFlags ` and if `flagColour` is anything else than any it knows that color might not be provided.

PR #292, opened 2021-01-08
spike: dynamic front pages

c-j-j:

### DIFF
--- a/packages/slice-layout/src/templates/frontleadtwo/index.js
+++ b/packages/slice-layout/src/templates/frontleadtwo/index.js
@@ -6,30 +6,54 @@ import HorizontalLayout from "../horizontallayout";
 import { getStyles } from "./styles";
 import { useResponsiveContext } from "@times-components-native/responsive";
 
+const baseWindowWidth = 1024;
+
+const baseWindowHeight = 768;
+const baseContentWidth = baseWindowWidth - 40 * 2;
+const baseContentHeight = 520;
+const baseAspectRatio = baseContentWidth / baseContentHeight;
+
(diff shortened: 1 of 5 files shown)


packages/styleguide/src/breakpoints/index.js
```diff
@@ -40,18 +40,18 @@ const getNarrowArticleBreakpoint = (width) => {
 };
 
 const getStyleByDeviceSize = (styles, windowWidth, windowHeight) => {
-  const selectedSize = Object.entries(styles)
-    .sort(([a], [b]) => b - a)
-    .find(([size]) => size <= windowWidth);
+  const sortedStyles = Object.entries(styles).sort(([a], [b]) => b - a);
```
c-j-j, line 43: This is a temporary fix to address issues on android nexus 7 - we currently don't have a 960px width in our styles, so this just assigns that size the 1024px style.

packages/slice-layout/src/templates/frontleadtwo/index.js
```diff
@@ -6,30 +6,54 @@ import HorizontalLayout from "../horizontallayout";
...
   const { windowWidth, windowHeight } = useResponsiveContext();
   const styles = getStyles(orientation, windowWidth, windowHeight);
 
+  const heightAvailable = windowHeight - 248;
+
+  const currentAspectRatio = windowWidth / windowHeight;
+  const aspectRatioToUse = findClosestAspectRatio(currentAspectRatio);
+
+  const width = heightAvailable * aspectRatioToUse;
+  const multiplier = heightAvailable / baseContentHeight;
```
c-j-j, line 33: This `multiplier` is the main value required to scale all text sizes by a particular amount.

PR #293, opened 2021-01-11
fix: (TNLT-6168) - Android - display carousel in modal 

kamiranoff:
- Display carousel in Android if multiple images in the article

Before:
Screenshot_1610389152

After:
Screenshot_1610389255
Screenshot_1610472473


packages/article-lead-asset/src/article-lead-asset.js
```diff
@@ -71,7 +70,7 @@ const ArticleLeadAsset = ({
     <View style={style}>
       <LeadAsset
         aspectRatio={getRatio(crop.ratio)}
-        caption={<Caption {...caption} />}
+        caption={caption}
```
kamiranoff, line 73: This is not as flexible as before but `Caption` was the only component used throughout the project.

I could also add the possibility back to put a component, which will result in fewer test changes.

packages/image/src/modal-image/modal-image.js
```diff
@@ -32,21 +33,38 @@ class ModalImage extends Component {
...
-    return React.cloneElement(caption, { style });
+    if (this.state.currentIndex === 0) {
+      return (
+        <Caption style={style} credits={caption.credits} text={caption.text} />
+      );
+    }
+
+    if (
+      images[this.state.currentIndex - 1] &&
+      images[this.state.currentIndex - 1].attributes
```
c-j-j, line 56: You can use optional chaining here, i.e. `images[this.state.currentIndex - 1]?.attributes`

packages/image/src/modal-image/modal-image.js
```diff
@@ -32,21 +33,38 @@ class ModalImage extends Component {
...
-    const { caption } = this.props;
+    const { caption, images } = this.props;
     const style = isTablet ? tabletCaptionStyles : captionStyles;
 
     if (!caption) {
       return null;
     }
 
-    return React.cloneElement(caption, { style });
+    if (this.state.currentIndex === 0) {
```
c-j-j, line 48: Could destructure `currentIndex` out of `this.state` as you're using it in a few places.

packages/utils/src/get-extra-images-content.ts
```diff
@@ -0,0 +1,24 @@
+type Article = {
```
c-j-j, line 1: We have started to create some domain types here - so I think you can replace the `content` here with `ArticleContent` https://github.com/newsuk/times-components-native/blob/f4bdb911e7ffeca4ad3c95c89b319bec211f032a/packages/types/domain-types.ts#L75

packages/utils/src/get-extra-images-content.ts
```diff
@@ -0,0 +1,24 @@
...
+      relativeHeight: number;
+      relativeHorizontalOffset: number;
+      relativeVerticalOffset: number;
+      relativeWidth: number;
+      url: string;
+    };
+  }[];
+};
+
+export const getExtraImagesContent = (article: Article) => {
```
c-j-j, line 18: Minor naming suggestion: `getInlineImages` may be a bit simpler?

c-j-j (approved):
Looks good to me - only a few minor comments for you to peruse

packages/utils/src/get-extra-images-content.ts
kamiranoff: Ah yes. I chose `getExtraImagesContent` to show that it was returning more than just images.

packages/article-lead-asset/src/article-lead-asset.js
aronshamash: I think render props are a good thing but if only ever `Caption` this makes sense

packages/image/image.showcase.js
```diff
@@ -140,7 +140,7 @@ export default {
       component: () => (
         <View style={{ width: 300 }}>
           <Image
-            aspectRatio={1 / 1}
```
aronshamash, line 143: 😆

aronshamash (approved):
looks really good

c-j-j (approved):

PR #296, opened 2021-01-14
chore: refactors section utils

c-j-j:
1. Breaks down section utils file into separate files.
2. Breaks down unit tests to separate files
3. Converts to typescript (haven't properly created types for everything, but this is a good starting point)
4. We only publically export util functions that are used elsewhere (see https://github.com/newsuk/times-components-native/compare/chore/refactor-section-utils?expand=1#diff-952b758d8636f4ae4f12a4e498085e7b032948e84a7fd946fa529344b6d06eabR1)


justadamhopkins (approved):
LGTM!!

kamiranoff (approved):

PR #297, opened 2021-01-14
feat: adds variant for lead1and1 WIP

c-j-j:
- [ ] TileAA no longer used - can be removed
- [ ] TileAF is now a configurable tile, could potentially be renamed to something like TileDynamicVertical1


packages/section/src/section.tsx
```diff
@@ -134,7 +134,7 @@ const Section: React.FC<Props> = ({
 
   const data = isPuzzle
     ? createPuzzleData(slices, editionBreakpoint)
-    : prepareSlicesForRender(isTablet)(slices);
+    : prepareSlicesForRender(isTablet, title)(slices);
```
aronshamash, line 137: `sectionTitle` ?

packages/types/index.ts
```diff
@@ -1,2 +1,3 @@
+export * from "./ui-types";
```
aronshamash, line 1: 👍

packages/section/src/utils/addSliceConfig.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+interface TransformSlice {
+  index: number;
+  name: string;
+  sectionTitle: string;
+  transform: (slice: Slice) => Slice;
+}
+
+const transformLeadOneAndOneNews = {
```
aronshamash, line 16: `transformLeadOneAndOneNews` -> `leadOneAndOneNewsTransform` ?  🤷

packages/section/src/utils/addSliceConfig.ts
```diff
@@ -0,0 +1,42 @@
...
+  name: "LeadOneAndOneSlice",
+  transform: (slice: Slice) => ({
+    ...slice,
+    support: { ...slice.support, config: { showImage: true } },
+  }),
+};
+
+const sliceTransformations: TransformSlice[] = [transformLeadOneAndOneNews];
+
+export const addSliceConfig = (isTablet: boolean, sectionTitle: string) => (
```
aronshamash, line 28: was thinking `addSliceConfig` might be `transformSlices` ? @Geoff-Ford @justadamhopkins what do you think is clearer

packages/section/src/utils/addSliceConfig.ts
c-j-j: Yes that'll be better

packages/section/src/utils/addSliceConfig.ts
c-j-j: `transformSlices` is a fairly general name, and could result in logic being put here that isn't related to adding these `config` fields. Also, when this is called in the context of  `prepareSlicesForRender` - I think it helps describe what is happening here.

aronshamash (approved):

aronshamash (commented):
looks great

Geoff-Ford (approved):

PR #298, opened 2021-01-14
feat: (TNLT-6012) Trigger article read

aronshamash:
- https://nidigitalsolutions.jira.com/browse/TNLT-6012
- `onArticleRead` will be fired either on scroll of article page or after 6 seconds which ever comes first
- Native will filter calls coming from the in view article.


packages/article-skeleton/__tests__/shared.native.js
```diff
@@ -186,6 +190,29 @@ export default () => {
...
+    // {
+    //   name: "onArticleRead is called on scroll",
+    //   async test() {
+    //     const onArticleRead = jest.fn();
+
+    //     const testInstance = TestRenderer.create(
+    //       renderArticle({ onArticleRead }),
+    //     );
+
+    //     testInstance.root.findByType(ScrollView).props.onScroll();
```
aronshamash, line 202: I'm getting this 
```
    TypeError: Cannot read property 'nativeEvent' of undefined

      200 |         );
      201 | 
    > 202 |         testInstance.root.findByType(ScrollView).props.onScroll();
          |                                                        ^
      203 |         expect(onArticleRead).toHaveBeenCalled();
      204 |       },
      205 |     },
```

any ideas ?

packages/article-skeleton/__tests__/shared.native.js
```diff
@@ -186,6 +190,29 @@ export default () => {
...
+    // },
+    {
+      name: "onArticleRead is called after 6 seconds",
+      async test() {
+        const onArticleRead = jest.fn();
+        TestRenderer.create(renderArticle({ onArticleRead }));
+        await delay(6000);
+        expect(onArticleRead).toHaveBeenCalled();
+      },
+      timout: 6000,
```
aronshamash, line 214: Needed to override the default 5000 jest timeout.

aronshamash: I thought better than setting it in jest config

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -48,8 +49,28 @@ const ArticleWithContent = (props) => {
...
+  useEffect(() => {
+    if (!hasBeenRead) {
+      const delay = setTimeout(() => {
+        setArticleRead();
+      }, 6000);
+      return () => clearTimeout(delay);
+    }
+  }, [hasBeenRead]);
+
+  const onScroll = () => {
```
kamiranoff, line 70: Just as a note,
I like to name handler with `handle` (i.e `handleScroll`) It makes it clear it handles the event and leave `on...` for props that are passed through.
basically this naming pattern:
https://jaketrent.com/post/naming-event-handlers-react

packages/pages/src/article/article-base.js
```diff
@@ -71,6 +71,10 @@ const ArticleBase = ({
           error={omitErrors ? null : error}
           interactiveConfig={interactiveConfig}
           isLoading={isLoading || (omitErrors && error)}
+          onArticleRead={(id) => {
+            console.log(`onArticleRead with id: ${id}, fired`);
+            //onArticleRead
```
kamiranoff, line 76: is this not implemented?

aronshamash: no not yet. We have split the feature and the native part has not yet been developed. I can remove the console.log  and comment though so it just does nothing maybe ?

aronshamash: it will error until `nativeEvents.onArticleRead` exists.

kamiranoff: Ah maybe `// @TODO` but just nitpicking.

aronshamash: yes will update

kamiranoff (approved):

packages/article-skeleton/src/article-skeleton.js
aronshamash: Changed to : `scrollHandler` now

packages/article-skeleton/__tests__/shared.native.js
aronshamash: I used `jest.advanceTimersByTime(6000);`

c-j-j (approved):

PR #299, opened 2021-01-18
feat: (TNLT-6350) add 2 by 3 image fragment to support2 on leadtwonopicandtwo fragment

justadamhopkins:
Updated the `leadtwonopicandtwo` query to incorporate a 2:3 image ratio format from TPA for the support 2 section of the slice. This is needed as the new sports variant of this `leadtwonopicandtwo` requires 2:3 image format on portrait and 4:5 image for landscape.


packages/provider-queries/src/section_fragment.graphql
```diff
@@ -509,14 +509,17 @@ fragment commonSlicesSection on StandardSectionSlice {
       headline
       leadAsset {
         ...leadAsset32and45
+        ...leadAsset23
```
c-j-j, line 512: I didn't know you could do this with our image query. I assumed you'd have needed to add a `leadAsset32and45and23` fragment... this is much nicer. 👍

c-j-j (approved):

kamiranoff (approved):

packages/provider-queries/src/section_fragment.graphql
aronshamash: ah yes good spot

packages/provider-queries/src/section_fragment.graphql
```diff
@@ -482,9 +482,11 @@ fragment commonSlicesSection on StandardSectionSlice {
         ...baseArticleProps
         summary125: summary(maxCharCount: 125)
         summary300: summary(maxCharCount: 300)
+        summary1000: summary(maxCharCount: 1000)
       }
       teaser125: teaser(maxCharCount: 125)
       teaser300: teaser(maxCharCount: 300)
+      teaser1000: summary(maxCharCount: 1000)
```
c-j-j, line 489: should this be `teaser(maxCharCount: 1000)`?

justadamhopkins: ah annoying i added that other day

PR #301, opened 2021-01-18
fix: (TNLT-649) functional issues images in gallery

kamiranoff:
- https://nidigitalsolutions.jira.com/browse/TNLT-649
- fix the wrong ratio applied in the gallery
- Duplicated images in the gallery 
- wrong caption displayed in gallery.

Screenshot_1611145655
Screenshot_1611145679


packages/image/src/modal-image/modal-image.tsx
```diff
@@ -7,9 +7,9 @@ import Url from "url-parse";
 import CloseButton from "../close-button";
 import ModalCaptionContainer from "../modal-caption-container";
 import Image from "../image";
-import styles, { captionStyles, tabletCaptionStyles } from "../styles";
-import Caption from "@times-components-native/caption";
+import styles from "../styles";
 import { ImageContent } from "@times-components-native/types";
+import ModalCaption from "@times-components-native/image/src/modal-image/modal-caption";
```
kamiranoff, line 12: I need to fix this import

packages/image/src/modal-image/modal-image.tsx
```diff
@@ -7,9 +7,9 @@ import Url from "url-parse";
 import CloseButton from "../close-button";
```
kamiranoff, line 7: This package is using "react-native-image-zoom-viewer" which is a bit limiting. I wonder if we should create our own image viewer instead at some point

packages/image/src/modal-image/modal-image.tsx
```diff
@@ -1,15 +1,15 @@
 import React, { FC, Fragment, useState } from "react";
 import { ImageProps, Modal, SafeAreaView, StyleProp, View } from "react-native";
-import { ResponsiveContext } from "@times-components-native/responsive";
-import Button from "@times-components-native/link";
-import ImageViewer from "react-native-image-zoom-viewer";
 import Url from "url-parse";
+import ImageViewer from "react-native-image-zoom-viewer";
```
kamiranoff, line 4: This package is using "react-native-image-zoom-viewer" which is a bit limiting. I wonder if we should create our own image viewer instead at some point

packages/image/src/modal-image/modal-image.tsx
```diff
@@ -109,41 +84,38 @@ const ModalImage: FC<ModalImageProps> = ({
...
   const mainUrl = new Url(uri, true);
   mainUrl.query.offline = "true";
-  const urls = [{ url: mainUrl.toString() }].concat(
+  const urls = [
+    {
+      url: mainUrl.toString(),
+      credits: caption?.credits,
+      caption: caption?.text,
+      aspectRatio: aspectRatio || 1,
+      relativeHeight: relativeHeight,
```
c-j-j, line 93: minor point - this can be `relativeHeight, relativeWidth` and so on...

packages/image/src/modal-image/modal-image.tsx
```diff
@@ -109,41 +84,38 @@ const ModalImage: FC<ModalImageProps> = ({
...
     images
       .map((i) => {
         const offlineUrl = new Url(i.attributes.url, true);
         offlineUrl.query.offline = "true";
-        return { url: offlineUrl.toString() };
+        const [ratioWidth, ratioHeight] = i.attributes.ratio.split(":");
+
+        return {
+          url: offlineUrl.toString(),
+          credits: i.attributes.credits,
```
c-j-j, line 107: You can remove a few lines here with the following:

```
return {...i.attributes, 
        url: offlineUrl.toString(), 
        aspectRatio: Number(ratioWidth) / Number(ratioHeight) || 1 }
```

c-j-j (approved):
Looks good - left some minor optional comments

packages/image/src/modal-image/modal-image.tsx
aronshamash: what are the things that it can't do that we would like ?

packages/image/src/modal-image/modal-image.tsx
```diff
@@ -109,41 +84,36 @@ const ModalImage: FC<ModalImageProps> = ({
...
+      caption: caption?.text,
+      aspectRatio: aspectRatio || 1,
+      relativeHeight,
+      relativeWidth,
+      relativeHorizontalOffset,
+      relativeVerticalOffset,
+    },
+  ].concat(
     images
       .map((i) => {
```
aronshamash, line 100: would prefer `i` -> `image`

packages/image/src/modal-image/modal-image.tsx
```diff
@@ -109,41 +84,36 @@ const ModalImage: FC<ModalImageProps> = ({
...
         const offlineUrl = new Url(i.attributes.url, true);
         offlineUrl.query.offline = "true";
-        return { url: offlineUrl.toString() };
+        const [ratioWidth, ratioHeight] = i.attributes.ratio.split(":");
```
aronshamash, line 103: we have this in various places now. Maybe we move to `image/utils`, main `utils` "package" or each package has its own? @c-j-j . My PR has created a helper see: https://github.com/newsuk/times-components-native/pull/305/files because we need in edition slices.

aronshamash (commented):
looks good in general.

packages/image/src/modal-image/modal-image.tsx
c-j-j: I think a shared util across the project would be good. `image/utils` sounds sensible

kamiranoff: should I create a new `utils` folder under `packages/image` or do you have an example of the pattern I should follow?

packages/image/src/modal-image/modal-image.tsx
kamiranoff: I had to add this property: `saveToLocalByLongPress={false}` because a `save to album` modal was displaying behind the caption 
That is modal that was showing behind the caption.
Screenshot_20210120_131039_uk co thetimes

there is small flicker I wasn't able to completely remove while swiping through images the first time. (But that might not be related to the imageViewer, more than the fact that the placeholder shows first).

Also (might be related to debug js remotely), it seemed to me that sometimes `renderImage` prop was not called after swiping.

But it works pretty well overall. Maybe it is just my frustration talking...

aronshamash: ok thanks for the reply

packages/image/src/modal-image/modal-image.tsx
aronshamash: we can leave as it is as you need to get it merge but I will create the helpers in `image/utils` and we can then update

aronshamash (approved):

PR #302, opened 2021-01-19
chore: first flexible tile, and refactoring tile logic to TS

c-j-j:
- [ ] TileAF can be deleted - to be done in a separate PR
- [ ] We can remove the `url` from the Link component as it's not used (TODO comment left in this PR). This can be done in another PR as it'll change 100s of snapshots.


packages/edition-slices/src/flexible-tiles/tile-vertical-a/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  TileImage,
+  TileLink,
+  TileSummary,
+  withTileTracking,
+} from "../../tiles/shared";
+
+import styleFactory from "./styles";
+import WithoutWhiteSpace from "../../tiles/shared/without-white-space";
+import PositionedTileStar from "../../tiles/shared/positioned-tile-star";
+import { TransformedTile } from "@times-components-native/types";
```
aronshamash, line 14: `ConfiguredTile`  any better you think ?

aronshamash, line 14: `ConfiguredTile`  any better you think ?

packages/edition-slices/__tests__/ios/tile-vertical-a.test.js
```diff
@@ -0,0 +1,3 @@
+import shared from "../tile-vertical-a/shared-tile-vertical-a.native";
```
kamiranoff, line 1: do we still the .native file?

aronshamash: no we shouldn't

packages/edition-slices/src/flexible-tiles/tile-vertical-a/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+interface Props {
+  onPress: () => void;
```
kamiranoff, line 17: shouldn't that be `onPress: (args: { id: string }) => void;` as defined in `tile-link`?
or  maybe `TileLinkProps['onPress']`

packages/edition-slices/__tests__/ios/tile-vertical-a.test.js
c-j-j: Good spot... we don't need the .native.js extension here. 

However, we do have around 230 files with that extension. Is it worth creating a new PR that'll change all of them, including this? Could cause a few conflicts

aronshamash: I think that is for a separate PR to update all of them

packages/edition-slices/src/flexible-tiles/tile-vertical-a/styles.tsx
```diff
@@ -0,0 +1,32 @@
+import {
+  fonts,
+  spacing,
+  editionBreakpoints,
+  globalSpacingStyles,
+} from "@times-components-native/styleguide";
+
+const fontSizeResolver: Record<string, number> = {
```
kamiranoff, line 8: I think the type could be improved with : 
`Record<keyof typeof editionBreakpoints, number>`

Not entirely sure

packages/edition-slices/src/flexible-tiles/tile-vertical-a/styles.tsx
```diff
@@ -0,0 +1,32 @@
...
+  [editionBreakpoints.medium]: 20,
+  [editionBreakpoints.wide]: 20,
+  [editionBreakpoints.huge]: 22,
+};
+
+export default (breakpoint: string) => ({
```
kamiranoff, line 14: Just curious,
There is an argument against `default export`:
https://basarat.gitbook.io/typescript/main-1/defaultisbad

I think it is debatable depending on what the team think

packages/edition-slices/src/tiles/shared/tile-image.tsx
```diff
@@ -1,16 +1,29 @@
...
+  relativeHeight: number;
+  relativeVerticalOffset: number;
+  relativeHorizontalOffset: number;
+  fill: boolean;
+  style: ViewStyle;
+  onLayout?: () => void;
+  uri: string;
+}
+
+const TileImage: FC<Props> = ({ hasVideo, style, onLayout, ...props }) => {
```
kamiranoff, line 20: Also not related to this PR, but the prop drilling `...props`. I wonder if it wouldn't be better to be explicit about the props passed in.

kamiranoff (approved):

packages/edition-slices/src/flexible-tiles/tile-vertical-a/index.tsx
c-j-j: Yes, good point. I think it ultimately calls this function here https://github.com/newsuk/times-components-native/blob/f0aba7c4fbbf30e7f66cf9c0d5d5ca0ce785515f/packages/pages/src/section/section.js#L27

So I think I'll create a shared type between these two.

aronshamash (approved):
just with the dir change. Lets tackle the other stuff as we go along

packages/edition-slices/src/flexible-tiles/tile-vertical-a/styles.tsx
c-j-j: I totally agree. I do prefer named exports, and definitely think they make a codebase easier to work with. However, this repo predominantly uses default exports, so it's going to be a big task to change/will introduce inconsistency

packages/edition-slices/src/tiles/shared/tile-image.tsx
c-j-j: I agree with the point, this was really just converting a js file to tsx to fix some typescript issues without any other changes. I'm happy to create a subsequent PR that defines what types the rest of those `props` have.

packages/edition-slices/src/flexible-tiles/tile-vertical-a/styles.tsx
c-j-j: A lot the styleguide code isn't currently typechecked, including the `eidtionBreakpoints` and where they're defined. Would be nice to do something like you've when that is converted over, and maybe creating a shared type as a lot of places have this breakpoint->style mapping

packages/edition-slices/src/tiles/shared/tile-image.tsx
aronshamash: definitely lets go with this approach in future PRs

PR #304, opened 2021-01-19
chore: removes unused Link "url" prop

c-j-j:
The `Link` component doesn't accept a `url` prop (see https://github.com/newsuk/times-components-native/blob/7765f923075cf5fcdaf6f9558c383177ad39406d/packages/link/src/link.android.js#L5 and https://github.com/newsuk/times-components-native/blob/7765f923075cf5fcdaf6f9558c383177ad39406d/packages/link/src/link.js#L5), however `TileLink` was providing this anyway.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -51,7 +51,6 @@ module.exports = {
     "jsx-a11y/anchor-is-valid": [
       "error",
       {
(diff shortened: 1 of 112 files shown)


kamiranoff (approved):
Nice cleanup!

PR #306, opened 2021-01-20
chore: renames .native.js files to .js

c-j-j:
As this project is only for native components, it's unnecessary to have `.native.js` files anymore.

Bash commands to do this were:

## Rename all files ending with `.native.js`
```
find . -name "*.native.js" | sed -e "p;s/native\.js/js/" | xargs -n2 git mv
```

## Remove `.native` from imports

```
grep -rl "import.*\.native\b" . | xargs sed -i "" -e 's/\.native//g'
```

All files changed were either in the `__tests__` directory or storybook showcase files


aronshamash (approved):

kamiranoff (approved):
I thought it would result in more deleted files?

Anyway, very nice!

c-j-j:
> I thought it would result in more deleted files?
> 
> Anyway, very nice!

@kamiranoff unfortunately not. Every `.native.js` file imports a shared.js file - which is really unnecessary now but also isn't something we can quickly address (at least as far as i know)

PR #309, opened 2021-01-20
fix(TNLT-6347): changes front lead2 portrait grid

c-j-j:
https://nidigitalsolutions.jira.com/browse/TNLT-6347

# Amendments to Dynamic Front Tile Logic
Previously, our front tiles rendered the following in order of priority, and anything that doesn't fit simply was removed from the tile `headline/strapline (if present) + byline + content`

This exposed an issue with the front lead2 when the design changed - whereby we were rendering headline+byline, and then with a large space underneath as there's not enough room for content. To better fill the tile in this context, it's better to render headline+content. So the new rules are as follows:

1. If there is not enough room for headline+byline+content, but there is room for headline+content, then we show headline+content
2. If there is not enough room for headline+byline+content, and there’s only room for headline+byline, then we show headline+byline
3. If there is room for headline+byline+content, then show headline+byline+content

You can see the impact of this change in the Lead Two screenshots with the 1-deck and 2-deck headlines.

# Lead Two 
## Before - iPad (8th gen 768x1024)
Simulator Screen Shot - iPad (8th generation) - 2021-01-20 at 13 55 44

## After - iPad (8th gen 768x1024)
Simulator Screen Shot - iPad (8th generation) - 2021-01-21 at 13 09 14
Simulator Screen Shot - iPad (8th generation) - 2021-01-21 at 13 09 01

# Lead One And One
## Before - iPad (8th gen 768x1024)
image

## After - iPad (8th gen 768x1024)
Simulator Screen Shot - iPad (8th generation) - 2021-01-21 at 13 08 41

### DIFF
--- a/packages/front-page/utils/get-front-tile-config.ts
+++ b/packages/front-page/utils/get-front-tile-config.ts
@@ -22,18 +22,18 @@ interface SummaryConfig {
   };
 }
 
-function isLastItem(items: string[], item: string) {
+const shouldRemoveMarginBottom = (items: string[], item: string) => {
   const pos = items.indexOf(item);
 
   if (pos < 0) return true;
 
   return pos === items.length - 1;
-}
+};
 
-function getNumberOfLines(heightAvailable: number, lineHeight: number) {
+const getNumberOfLines = (heightAvailable: number, lineHeight: number) => {
   if (heightAvailable < 0) return 0;
   return Math.floor(heightAvailable / lineHeight);
-}
+};
 
 export const getFrontTileConfig = (summaryConfig: SummaryConfig) => {
   const { container, headline, strapline, bylines, content } = summaryConfig;
@@ -88,17 +88,19 @@ export const getFrontTileConfig = (summaryConfig: SummaryConfig) => {
   return {
     headline: {
       show: true,
-      marginBottom: isLastItem(itemsToRender, "headline") ? 0 : headlineMargin,
+      marginBottom: shouldRemoveMarginBottom(itemsToRender, "headline")
+        ? 0
+        : headlineMargin,
     },
     strapline: {
       show: itemsToRender.includes("strapline"),
-      marginBottom: isLastItem(itemsToRender, "strapline")
+      marginBottom: shouldRemoveMarginBottom(itemsToRender, "strapline")
         ? 0
         : strapline.marginBottom,
     },
     byline: {
       show: itemsToRender.includes("byline"),
-      marginBottom: isLastItem(itemsToRender, "byline")
+      marginBottom: shouldRemoveMarginBottom(itemsToRender, "byline")
         ? 0
         : bylines.marginBottom,
     },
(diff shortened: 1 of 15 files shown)


packages/front-page/utils/get-front-tile-config.ts
```diff
@@ -43,37 +43,68 @@ export const getFrontTileConfig = (summaryConfig: SummaryConfig) => {
...
+      marginBottom: canAccommodateContentWithByline ? bylines.marginBottom : 0,
     },
+    content: canAccommodateContentWithByline
+      ? {
+          show: true,
+          marginBottom: 0,
+          numberOfLines: Math.floor(
+            heightForContentWithByline / content.lineHeight,
+          ),
+        }
```
kamiranoff, line 95: a small thing, I guess you could add this to a variable as it is duplicated.

packages/front-page/utils/get-front-tile-config.ts
```diff
@@ -43,37 +43,68 @@ export const getFrontTileConfig = (summaryConfig: SummaryConfig) => {
...
+    !canAccommodateContentWithByline && canAccommodateContentWithoutByline;
 
   return {
     headline: {
       show: true,
       marginBottom:
-        canAccommodateStrapline || canAccommodateByline || canAccommodateContent
+        canAccommodateStrapline ||
+        canAccommodateByline ||
+        canAccommodateContentWithoutByline
```
kamiranoff, line 70: I wonder if a function `getHeadlineMargin` would make it more readable. 
But I can not much that can be done with so many business logic rules.

kamiranoff (approved):
Just commented for the sake of it. Nice one.

packages/front-page/utils/get-front-tile-config.ts
c-j-j: The `numberOfLines` is ever so slightly different. I could add a helper function that takes `heightForContentWithByline`/`heightForContentWithoutByline` that'll help simplify this

```
          numberOfLines: Math.floor(
            heightForContentWithByline / content.lineHeight,
          ),
```
```
          numberOfLines: Math.floor(
            heightForContentWithoutByline / content.lineHeight,
          ),
          ```

File: packages/front-page/utils/get-front-tile-config.ts (hunk 2 above)
Comment (c-j-j) [reply packages/front-page/utils/get-front-tile-config.ts:70]: Yes I'm not happy with how this is looking. It's getting very complicated. I'll have a play and see if I can help simplify some of this complexity. Maybe something like this will be a bit easier to read?

```
  const itemsToRender = [
    "headline",
    canAccommodateStrapline && "strapline",
    canAccommodateByline && "byline",
    canAccommodateContentWithByline && "content",
    canAccommodateContentWithoutByline && "content",
  ].filter(Boolean);

  return {
    headline: {
      show: true,
      marginBottom: isLastItem(itemsToRender, "headline") ? 0 : headlineMargin,
    },
```

kamiranoff: ah yes. not bad!

c-j-j: Pushed a change... Seems to be working quite nicely

aronshamash:
looks good.

PR #311, opened 2021-01-20
chore: Migrate drop cap to new approach

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-5966

In essence I have made a new way of creating drop caps using the recent InlineContent approach. I have removed all references to the previous J/absolute positioning approach and also other previous drop cap approaches I have found scattered around the codebase. I won’t add screenshots as this should basically look exactly the same as before and there are many, many variants of fonts\templates\devices\scale etc.

Sorry it is so big. Wasn't really possible to break this up into smaller chunks.

I had to stop myself ripping out all the remaining absolute positioning logic and AttributedString stuff which shouldn't now be used anywhere just to keep this PR at something reasonable. But we really do need to do this otherwise that code will confuse.

Also worth mentioning is that we are pre-measuring the width of the drop cap in the setup stage. This could potentially be done in the render measuring stage of InlineContent which is already measuring height. But again, I didn't want this PR to touch any more code so a possible one for the future.


packages/drop-cap/src/drop-cap.tsx
```diff
@@ -0,0 +1,72 @@
...
+const DropCap: React.FC<Props> = ({
+  dropCapColor,
+  dropCapFont,
+  dropCapFontSize,
+  dropCapText,
+  height,
+  scale,
+  width,
+}) => {
+  const baseAdditionalPadding = Platform.OS === "ios" ? 23 : 20;
```
c-j-j, line 37: Did you look at setting `includeFontPadding` to false in the `Text` component? Android adds some additional padding - which can be disabled this way. Maybe that'll mean we can use same `baseAdditionalPadding` for both platforms..

c-j-j (approved):
All looks good as far as i can see. I tested on iPad pro 12 and everything seemed to work just fine.

packages/drop-cap/src/drop-cap.tsx
Geoff-Ford: I did try that yes but it didn't make any difference.

packages/article-skeleton/src/body-utils/setupInlineContent.js
```diff
@@ -0,0 +1,79 @@
+export const setupInlineContent = (
+  skeletonProps,
+  unprocessedContent,
+  processedContent = [],
+) => {
+  const { isTablet } = skeletonProps;
+  if (!isTablet) return unprocessedContent;
+
+  const numberOfCandidateParagraphsToInline = 7;
```
kamiranoff, line 9: out of curiosity, where does this 7 come from?

Geoff-Ford: This isn't new code btw, I just broke out what was becoming a large file into separate files. But yeah, the 7 comes from trial an error. The paragraphs to be inlined can of course be different sizes but we needed to ensure that there was enough content to wrap around say a double MPU adverts. Any of those 7 that don't need to be inlined (after measuring we know this) are rendered normally below.

aronshamash (approved):
good work. if there is any more tweaking I think it is minor and can be done in future fix prs

kamiranoff (approved):

justadamhopkins (approved):
LGTM! I verified on a few simulator devices for both android and ios as the pr was so large.

PR #313, opened 2021-01-22
feat: (TNLT-6336/TNLT-6352) secondary-four right-aligned variants (tablet only)

c-j-j:
This PR ensures that a secondary-four slice displays right-aligned images after they follow a slice with left-aligned images.

There were a few issues with the previous implementation which are now fixed: 
1. We weren't alternating the image alignment from left-right-left and so on. We were only doing left-right-right-etc
2. When we were flipping a secondary-four slice to be right-aligned, the articles were also being flipped, which isn't correct. The articles need to appear in the same order in the slice from left-to-right but have a different display.

# TopSecondarySlice / Secondary Four
https://nidigitalsolutions.jira.com/browse/TNLT-6336

## Before
image
## After
simulator_screenshot_A3085C48-2BD1-4A15-A2E8-7DEA9234A8A2

# LeadOneAndOne / Secondary Four
https://nidigitalsolutions.jira.com/browse/TNLT-6352

## Before
image

## After
simulator_screenshot_ADD1C73B-C607-43DA-9C9B-8DC596E09CF0

### DIFF
--- a/packages/section/src/utils/consecutiveItemsFlagger.ts
+++ b/packages/section/src/utils/consecutiveItemsFlagger.ts
@@ -2,7 +2,6 @@ import memoizeOne from "memoize-one";
 
 const withIsConsecutive = (slice: any) => ({ ...slice, isConsecutive: true });
 
-// in order to
 const consecutivePairs = [
   ["TopSecondaryFourSlice", "SecondaryFourSlice"],
   ["LeadOneAndOneSlice", "SecondaryFourSlice"],
--- a/packages/section/src/utils/consecutiveItemsFlagger.ts
+++ b/packages/section/src/utils/consecutiveItemsFlagger.ts
@@ -1,8 +1,11 @@
 import memoizeOne from "memoize-one";
+import { SliceName } from "@times-components-native/types";
 
 const withIsConsecutive = (slice: any) => ({ ...slice, isConsecutive: true });
 
-const consecutivePairs = [
+type Pair = [SliceName, SliceName];
+
+const consecutivePairs: Pair[] = [
   ["TopSecondaryFourSlice", "SecondaryFourSlice"],
   ["LeadOneAndOneSlice", "SecondaryFourSlice"],
 ];
--- a/packages/section/src/utils/consecutiveItemsFlagger.ts
+++ b/packages/section/src/utils/consecutiveItemsFlagger.ts
@@ -2,15 +2,30 @@ import memoizeOne from "memoize-one";
 
 const withIsConsecutive = (slice: any) => ({ ...slice, isConsecutive: true });
 
+// in order to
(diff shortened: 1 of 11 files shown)


packages/section/src/utils/consecutiveItemsFlagger.ts
```diff
@@ -2,15 +2,29 @@ import memoizeOne from "memoize-one";
 
 const withIsConsecutive = (slice: any) => ({ ...slice, isConsecutive: true });
 
+const consecutivePairs = [
+  ["TopSecondaryFourSlice", "SecondaryFourSlice"],
+  ["LeadOneAndOneSlice", "SecondaryFourSlice"],
```
kamiranoff, line 7: would be good to have these values in an enum I think

kamiranoff (approved):

kamiranoff:
One small comment.

packages/section/src/utils/consecutiveItemsFlagger.ts
c-j-j: Good suggestion... although might make sense to do that in a separate PR? We have a few places that we've converted over to typescript assuming that a slice name is just a string at the moment.

kamiranoff: yes, make sense!

c-j-j: I think a union type (e.g. `type SliceName = 'LeadOneAndOneSlice | 'AnotherSlice'` etc) may make things easier than an enum for now. Do you think that'll suffice?

We need to review how we're using typescript for things like the edition tree/slices etc - as these things are largely untyped. We do have some generated types from the TPA schema, although we've struggled to use them in the past as they're missing info for things like article content - so we've had to create our own

c-j-j: I think ideally we would have a step whereby when we receive a edition tree as JSON - we would convert this to a typed object, including converting strings to enums, but we're not quite at that stage to do that yet.

c-j-j: @kamiranoff I changed my mind... I added a SliceName type as part of this PR and used it here.

aronshamash: looks good

packages/types/domain-types.ts
```diff
@@ -90,3 +90,28 @@ export type ArticleContent =
   | InteractiveContent
   | ImageContent
   | AdContent;
+
+export type SliceName =
```
aronshamash, line 94: @justadamhopkins we can use this in our PR when this is merged

aronshamash (approved):

PR #319, opened 2021-01-27
feat: (TNLT-6359/TNLT-6292/TNLT-6334) lead magazine slice

c-j-j:
- https://nidigitalsolutions.jira.com/browse/TNLT-6359
- https://nidigitalsolutions.jira.com/browse/TNLT-6292
- https://nidigitalsolutions.jira.com/browse/TNLT-6334

## Noteworthy changes
1. This new slice is for iOS only. We fallback to the existing magazine slice on android.
2. iOS passes a new contentSize initial prop to us, which is the size of the view that RN is rendered into. This isn't strictly necessary to get this slice working on iOS, but will be when we're porting this over to android. This explains why there are quite a few other snapshot changes in this PR.
3. The first tile in the slice takes up 100% width, which means we needed to structure this slice differently to all others by only wrapping the support tiles in a `Responsive` container instead of the whole slice.

#### iPad Mini 4
Screenshot 2021-02-01 at 16 14 02

#### iPad Air
Screenshot 2021-02-01 at 16 10 30

#### iPad Mini 4
Simulator Screen Shot - iPad mini 4 - 2021-02-02 at 15 29 37
Simulator Screen Shot - iPad mini 4 - 2021-02-02 at 15 33 01

#### iPad Pro 12''
Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-02-02 at 15 25 19
Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-02-02 at 15 25 26


packages/edition-slices/src/tiles/tile-lead-supplement-portrait/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+        relativeWidth={crop.relativeWidth}
+        relativeHeight={crop.relativeHeight}
+        relativeHorizontalOffset={crop.relativeHorizontalOffset}
+        relativeVerticalOffset={crop.relativeVerticalOffset}
+        style={styles.imageContainer}
+        uri={crop.url}
+        hasVideo={hasVideo}
+      />
+      <TileSummary
+        headlineStyle={styles.headline}
```
c-j-j, line 57: Do you need to pass labelColor? I added this https://github.com/newsuk/times-components-native/pull/319/files#diff-8dff6d7ba109a162e8c886f615e10262f56ad292ff9f9ed53f806ad88471a082R61

aronshamash: Waiting for UXD to confirm requirements

aronshamash: assuming same as existing tiles which do not take a `labelColor`

packages/article-summary/src/article-summary.js
```diff
@@ -130,7 +130,11 @@ ArticleSummary.propTypes = {
   }),
   saveStar: PropTypes.node,
   strapline: PropTypes.node,
-  style: PropTypes.shape({}),
+  style: PropTypes.oneOfType([
```
kamiranoff, line 133: From memory, the propType is `number` if you use StyleSheet.create

packages/edition-slices/src/edition-slices.js
```diff
@@ -22,23 +22,29 @@ import {
...
   SupplementSecondaryFourSlice,
   SupplementSecondaryOneSlice,
   SupplementSecondaryTwoAndTwoSlice,
 } from "./slices";
 import { sectionConfigs } from "@times-components-native/section/src/utils";
+import { Platform } from "react-native";
 
 const { sectionTitles } = sectionConfigs;
 
+const isIOS = Platform.OS === "ios";
```
kamiranoff, line 36: Not related to this PR but would it be useful to have this accessible globally?

packages/edition-slices/src/tiles/shared/tile-link.tsx
```diff
@@ -1,15 +1,13 @@
...
-} from "@times-components-native/types";
+import { ViewStyle } from "react-native";
+import { OnArticlePress } from "@times-components-native/types";
+import { Tile } from "@times-components-native/fixture-generator/src/types";
 
 interface Props {
   onPress: OnArticlePress;
-  style?: TextStyle;
-  tile: TransformConfiguredTile;
+  style?: ViewStyle;
```
kamiranoff, line 9: I thought that was the type for it `StyleProp<ViewStyle>`? Does it work as well?

kamiranoff (approved):

packages/article-summary/src/article-summary.js
c-j-j: This is a `style` object - not sure I understand why it'll be `number`? It can be falsy, which is why I've had to add a boolean as a potential type.

packages/edition-slices/src/edition-slices.js
c-j-j: Yes that would be useful as it is duplicated in quite a few places. But best to do in another PR as this is already getting pretty big

packages/edition-slices/src/tiles/shared/tile-link.tsx
c-j-j: Ah you're right, ViewStyle is a single style object, whereas StyleProp can allow array of objects/false/etc

packages/article-summary/src/article-summary.js
c-j-j: I think article-summary really needs to be converted to TS soon... where we can type these props correctly.

packages/edition-slices/__tests__/tile-lead-supplement-landscape/tile-lead-supplement-landscape.base.js
```diff
@@ -0,0 +1,20 @@
...
+
+    it("wide", () => {
+      testTile(TileLeadSupplementLandscape, editionBreakpoints.wide);
+    });
+
+    it("huge", () => {
+      testTile(TileLeadSupplementLandscape, editionBreakpoints.huge);
+    });
+  });
+};
```
justadamhopkins, line 20: i know this is the consistent pattern but a `test.each` would keep these files leaner

packages/edition-slices/src/edition-slices.js
```diff
@@ -22,23 +22,29 @@ import {
...
   const isInTabletSupplement = isInSupplement && isTablet;
   return {
     CommentLeadAndCartoonSlice,
     DailyUniversalRegister: DailyRegisterLeadFourSlice,
     LeadersSlice,
     LeadOneAndFourSlice: isInTabletSupplement
-      ? SupplementLeadOneAndFourSlice
+      ? isIOS
+        ? SupplementLeadOneAndFourV2Slice
+        : SupplementLeadOneAndFourSlice
```
justadamhopkins, line 47: Should you not break this out into a function below? like I did with the l2np2 sport variant?

justadamhopkins (changes requested):
Looks good one minor comment on consistency in the slice map

packages/edition-slices/__tests__/tile-lead-supplement-landscape/tile-lead-supplement-landscape.base.js
c-j-j: I do like `test.each` - but as you say it, will introduce inconsistency adding here. Potentially something we can do in a separate PR?

packages/edition-slices/src/edition-slices.js
c-j-j: Good call - I can do the others too so this main function reads a bit better.

justadamhopkins (approved):
LGTM!!

PR #320, opened 2021-01-27
refactor(TNLT-6524): hide close button on android for tooltip

justadamhopkins:
Temporarily hid the tooltip close button on android due to underlying functionality issue.

Screenshot 2021-01-28 at 12 23 03
Screenshot 2021-01-28 at 12 25 53


packages/tooltip/__tests__/shared.js
```diff
@@ -219,5 +220,24 @@ export default () => {
       closeButton.simulate("press");
       expect(onCloseMock).toBeCalled();
     });
+
+    it("hides tooltip close button on android only", () => {
```
kamiranoff, line 224: shouldn't this test be under the android folder?

justadamhopkins: yes good spot and it made me realised I needed to split the tests out more as some assert for the existence of the close button which would now only work on ios.

kamiranoff (approved):

PR #323, opened 2021-01-29
fix: TNLT-6532 opta widget showing error on mobile

kamiranoff:
- https://nidigitalsolutions.jira.com/secure/RapidBoard.jspa?rapidView=1334&projectKey=TNLT&selectedIssue=TNLT-6532
- fix Opta widget showing error on android

**Before:**
Screenshot_20210129_130409_uk co thetimes debug

**After:**
Screenshot_20210129_130340_uk co thetimes debug


packages/interactive-wrapper/src/interactive-wrapper.android.js
```diff
@@ -20,10 +20,12 @@ class InteractiveWrapper extends Component {
...
   }
 
-  constructor() {
-    super();
+  constructor(props) {
+    super(props);
     this.state = {
       height: 1,
+      hasError: false,
+      hasLoaded: false,
```
Geoff-Ford, line 28: Not a biggie but we generally use `isLoading` in other parts of TCN and default it to `true` then set it to `false` when loading complete.

kamiranoff: I'll change it. Will make the if statement easier to read as well

packages/interactive-wrapper/src/interactive-wrapper.android.js
```diff
@@ -62,9 +71,13 @@ class InteractiveWrapper extends Component {
       config: { dev, environment, platform, version },
       id,
     } = this.props;
-    const { height } = this.state;
+    const { height, hasError, hasLoaded } = this.state;
     const uri = `${editorialLambdaProtocol}${editorialLambdaOrigin}/${editorialLambdaSlug}/${id}?dev=${dev}&env=${environment}&platform=${platform}&version=${version}`;
 
+    if (hasError || !hasLoaded) {
+      return null;
+    }
```
Geoff-Ford, line 79: This can all go on a single line.

Geoff-Ford (approved):
LGTM. Two minor points.

packages/interactive-wrapper/src/interactive-wrapper.android.js
kamiranoff: `if (hasError || isLoading) return null;` Prettier is happy with this if I remove the curly brackets.

Geoff-Ford: Yep, it's really a style thing but we have generally conformed to this if possible. If the conditional code is longer than 1 line then we keep it in a block.

packages/interactive-wrapper/src/interactive-wrapper.android.js
```diff
@@ -32,11 +34,18 @@ class InteractiveWrapper extends Component {
...
 
   onLoadEnd() {
+    this.setState({ isLoading: false });
     if (this.webview) {
       this.webview.postMessage("thetimes.co.uk", "*");
     }
   }
 
+  handleHttpError = (event) => {
+    if (event?.nativeEvent?.statusCode > 300) {
```
kamiranoff, line 44: Maybe that should be `400`. Not sure if we would ever get 300s?

c-j-j (approved):

aronshamash (approved):

PR #324, opened 2021-01-29
feat(TNLT-6530): pass article id to tooltip

justadamhopkins:
This change passes the article id to all instances of the tooltip component. The id is needed as a 2nd param in the `onTooltipPresentedMock` function only for android. 

The tooltip instances so far are 'profile, commenting and topics`.  Tests have also been updated to account for the new params being passed.

This implements a breaking change for the IOS bridge method which will need to be handled by UA team and Mobile team.


packages/tooltip/tooltip.tsx
```diff
@@ -12,11 +12,15 @@ interface Props {
   offsetX?: number;
   offsetY?: number;
   onClose?: <T = unknown, R = unknown>(args?: T) => R;
-  onTooltipPresented: <T = unknown, R = unknown>(args?: T) => R;
+  onTooltipPresented: <T = unknown, R = unknown>(
+    args: T,
+    articleId: string,
+  ) => R;
```
kamiranoff, line 18: It seems to me that; the type is a string and so is articleId and it doesn't return anything:
Wouldn't it be simpler to type it like this?:
```
onTooltipPresented: (type: string, articleId: string) => void;
  ```

kamiranoff (approved):

packages/tooltip/tooltip.tsx
justadamhopkins: yeah, i thought that but didn't want to change too much. I'll update.

justadamhopkins: Its changed

aronshamash (approved):
looks good.

PR #337, opened 2021-02-04
fix: (TNLT-6569) Lead 2 No Pic 2 styling on iPad Pro 12 inch

aronshamash:
- https://nidigitalsolutions.jira.com/browse/TNLT-6569
- https://nidigitalsolutions.jira.com/browse/TNLT-6574

### iPhone 11 

Screenshot 2021-02-04 at 16 06 57

### iPad Portrait

Screenshot 2021-02-04 at 16 07 32

### iPad Landscape
Screenshot 2021-02-04 at 16 08 29

### iPad Pro 12 Portrait
Screenshot 2021-02-04 at 15 37 50

### iPad Pro 12 Landscape
Screenshot 2021-02-04 at 15 38 18


packages/edition-slices/src/tiles/tile-d/styles/index.js
```diff
@@ -44,8 +44,14 @@ const mediumBreakpointStyles = {
...
     flex: 1,
     paddingLeft: spacing(2),
+    width: "100%",
   },
 };
 
-export default (breakpoint) =>
-  breakpoint === editionBreakpoints.medium ? mediumBreakpointStyles : styles;
+const stylesResolver = {
+  [editionBreakpoints.small]: mediumWideBreakpointStyles,
```
aronshamash, line 52: need to sort this

aronshamash: done

packages/edition-slices/src/tiles/tile-y/index.js
```diff
@@ -10,14 +10,24 @@ import {
...
+  tile,
+  breakpoint = editionBreakpoints.medium,
+  orientation,
+}) => {
+  const styles = stylesFactory(breakpoint, orientation);
+  const summaryLength =
+    (breakpoint === "wide" && orientation === "portrait") ||
+    breakpoint === "huge"
+      ? 1000
+      : 300;
```
c-j-j, line 24: I assume we're requesting both the 300 and 1000 summary lengths? I think we could have just got away with 1000 here, and it'll have truncated automatically on the smaller tile. Hopefully can prevent needing to do this once that spike is done on using the content as teaser.

c-j-j (approved):

kamiranoff (approved):

packages/edition-slices/src/tiles/tile-y/index.js
aronshamash: yes we already have `300` and `1000` in the query

aronshamash: it doesn't truncate automatically so left as it was. We can look at this as part of the spike?

c-j-j: yes, sure.

PR #342, opened 2021-02-05
feat: improve section performance

c-j-j:
1. Slices are currently re-rendering many times more than is needed. I converted all slices to PureComponent's, which stops the re-rendering 
2. The section flatlist doesn't need to only render 2 or 5 slices at a time. All the data is loaded into the app, and can be rendered all at once. This results in a better scrolling experience - and I didn't notice a massive jump in RAM when doing this.

On master, if I scroll down on a section, I can see the JS thread struggling to stay at 60fps and often dropping massively to 10/20fps. With this change, the JS thread stays at 60fps most of the time and hardly drops to the extent seen on master. 

There is still an issue with the UI thread whilst scrolling that I cannot figure out what is causing this. 

image

### DIFF
--- a/packages/edition-slices/src/slices/leaders/index.js
+++ b/packages/edition-slices/src/slices/leaders/index.js
@@ -1,4 +1,4 @@
-import React, { Component } from "react";
+import React, { PureComponent } from "react";
 import { View, Text } from "react-native";
 import { SectionContext } from "@times-components-native/context";
 import { Leaders } from "@times-components-native/slice-layout";
@@ -21,7 +21,7 @@ const renderHead = (styles, breakpoint) => (
   </SectionContext.Consumer>
 );
 
-class LeadersSlice extends Component {
+class LeadersSlice extends PureComponent {
   constructor(props) {
     super(props);
     this.renderSlice = this.renderSlice.bind(this);
--- a/packages/edition-slices/src/slices/commentleadandcartoon/index.js
+++ b/packages/edition-slices/src/slices/commentleadandcartoon/index.js
@@ -1,10 +1,10 @@
-import React, { Component } from "react";
+import React, { PureComponent } from "react";
 import { CommentLeadAndCartoon } from "@times-components-native/slice-layout";
 import PropTypes from "prop-types";
 import { TileP, TileQ, TileAH, TileAI } from "../../tiles";
 import { ResponsiveSlice } from "../shared";
 
-class CommentLeadAndCartoonSlice extends Component {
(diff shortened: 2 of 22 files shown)


aronshamash (approved):

aronshamash:
nice catch. I'll give it a go.

kamiranoff:
@c-j-j [why-did-you-render](https://github.com/welldone-software/why-did-you-render) and [react-profiler](https://reactjs.org/blog/2018/09/10/introducing-the-react-profiler.html) are good tools I know to debug slow performances in case you haven't tried. But I guess its never easy.

kamiranoff (approved):

packages/edition-slices/src/slices/leaders/index.js
```diff
@@ -21,7 +21,7 @@ const renderHead = (styles, breakpoint) => (
   </SectionContext.Consumer>
 );
 
-class LeadersSlice extends Component {
+class LeadersSlice extends PureComponent {
```
kamiranoff, line 24: Also. I have a question here. I have been trying to get my head around it.

As I understood `PureComponent` does a shallow comparison of the props. Meaning it will not re-render if the props are primitive values and equal and the same for arrays or objects if they only contain primitives value as the first level of depth. [shallowEqual.js](https://github.com/facebook/fbjs/blob/master/packages/fbjs/src/core/shallowEqual.js#L39-L67)

But complex objects will still trigger re-render.

I also believe if you pass a function as a prop, the PureComponent will always re-render, (unless, and that is an assumption, the function is wrapped in useCallback)

c-j-j: Interesting points... to check the unnecessary re-rendering, I added console.log statements in one of the slices that only appeared on the section once (i.e. LeadOneAndOne), and the regular `Component` re-rendered several times as I was scrolling down - whereas the PureComponent only rendered once. Which doesn't really make much sense based on what you've brought up about the shallow comparison - as the `slice` prop is not a shallow object and `shallowEqual` will return false. 

We should use those tools you've linked and properly investigate the section pages.

c-j-j: @kamiranoff Just looked into this a little further. For some reason, why-did-you-render didn't flag that the slices were being re-rendered unnecessarily. I logged out the props of a slice, and there were no changes between subsequent renders. Also I learnt that PureComponent's `shallowEqual` will return true when the object reference is the same. So I assume that because the object props are the same reference - it prevented a re-render

PR #343, opened 2021-02-05
chore: (TNLT-6479) article performance

c-j-j:
https://nidigitalsolutions.jira.com/browse/TNLT-6479

## Problem

When we're rendering an article, we are parsing the article AST into AttributedString objects. These AttributedString objects then get rendered by the SimpleParagraph component. These AttributedString objects were required by the previous inline logic, and are no longer required for how we render inline items now. Also, these AttributedString objects seem to be causing a delay when rendering an article (I'm seeing 20ms to render a single SImpleParagraph whereas this should be much faster)

## Solution

 We can render an article by just converting the AST into RN components.

### Noteworthy changes
1. New subscript/superscript renderers that just render smaller Text components 
2. Removed AttributedString/LayoutManager/SimpleParagraph/InlineParagraph as no longer used, plus few other files
3. unknown renderer now just returns the children
4. Refactored article-skeleton to not use useMemo/useCallback when rendering react components, and to instead render them as simple react components.


packages/article-skeleton/__tests__/ios/__snapshots__/article-header-with-style.ios.test.js.snap
```diff
@@ -70,30 +70,28 @@ exports[`1. article with header 1`] = `
...
-                selectable={true}
-              >
+              <Text>
                 last year that rote learning has “a lot to be said against it”, but added: “As the phrase ‘learning by heart’ implies, this sort of remembering places the poem at a central place in ourselves and makes it precious.”
               </Text>
-              <Text
-                selectable={true}
-              >
-                When Thatcher joined the campaign trail for the 2001 general election...
+              <Text>
```
kamiranoff, line 91: We might still need the `selectable={true}`  prop no?

c-j-j: I've added that to the `paragraph` renderer - so all text in a paragraph should now be selectable. Hopefully that's sufficient?

kamiranoff (approved):
Looks very good!

packages/article-skeleton/__tests__/ios/__snapshots__/article-header-with-style.ios.test.js.snap
c-j-j: Hmmm doesn't seem to work anymore - need to fix this. Good spot

kamiranoff: Yeah, I investigate an issue recently to allow selecting the text and individual words. It works out of the box in Android but couldn't make it work for IOS.
Here is the ticket:
https://nidigitalsolutions.jira.com/browse/TNLT-3837

kamiranoff: I believe have the wrapper Text component selectable is sufficient. But might be worth double checking

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -128,8 +113,13 @@ const ArticleWithContent = (props) => {
...
           scrollEventThrottle={400}
         >
-          {header}
-          {processedContent}
+          <Gutter>
+            <Header width={Math.min(maxWidth, windowWidth)} />
+          </Gutter>
+
+          {fixedContent.map((item, index) => (
+            <ContentChild key={index} item={item} index={index} />
```
kamiranoff, line 121: That might be dangerous to set the key as index:
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -123,55 +100,68 @@ export default ({
...
+      );
     },
     superscript(key, attributes, children) {
-      const childStr = AttributedString.join(children);
-      childStr.string = toSuperscript(childStr.string);
-      return childStr;
+      return (
+        <View style={{ flexDirection: "row", alignItems: "flex-end" }}>
+          <Text
+            style={{
```
kamiranoff, line 139: we might want to avoid setting online style. As well.
For performance too, as a new object will be recreated and trigger a new render even if children doesnt change. If this is in StyleSheet.Create Text will re-render only if `children` changes instead

c-j-j: That is true - all styles in this file should come from the shared stylesheet. There's quite a few renderers here that'll need to change - which makes sense to do as part of this PR.

packages/article-skeleton/src/article-skeleton.js
c-j-j: I'm not really sure what to use here instead really, as I don't think `fixedContent` has a unique id for me to use as a `key`. Also, `fixedContent` doesn't change afaik when it's rendered (i.e. isn't re-ordered or has new items added etc). If that is the case, is using the index acceptable?

kamiranoff: That is a good point. Might be just fine. I usually use a prefix when I don't have an id `fixedContent-${index}`.

packages/article-skeleton/__tests__/ios/__snapshots__/article-header-with-style.ios.test.js.snap
c-j-j: I think I'm seeing the same issue as on master. You can copy a whole paragraph but not individual words, and it doesn't highlight. As this isn't a regression, I will leave selectable on the paragraph and hopefully RN fix this at some point

packages/article-skeleton/src/article-skeleton.js
c-j-j: I can add a prefix. Won't entirely address the issues highlighted in that link you've shared, but at least it's more descriptive when debugging

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "3.0.0",
+  "version": "3.0.0-beta",
```
c-j-j, line 4: Needs removing before merging

packages/article-skeleton/__tests__/ios/__snapshots__/article-header-with-style.ios.test.js.snap
kamiranoff: Ah yes. That is the issue I couldn't fix on IOS but should work fine on android I think

packages/article-paragraph/src/index.js
```diff
@@ -4,38 +4,33 @@ import PropTypes from "prop-types";
...
-  uid,
   height,
   style,
+  ast,
   narrowContent,
   attributes,
 }) => {
-  const { children: astChildren } = ast;
-  if (!astChildren || astChildren.length === 0) {
+  if (!ast || ast.length === 0) {
```
Geoff-Ford, line 14: Can also be just `!ast.length`

packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -123,55 +100,68 @@ export default ({
...
+        <ArticleLink
+          url={href}
+          onPress={(e) => {
+            return onLinkPress(e, {
+              canonicalId,
+              type,
+              url: href,
+            });
+          }}
+        >
```
Geoff-Ford, line 116: Can this be done with implicit return?

```
onPress={(e) => onLinkPress(e, {
                               canonicalId,
                               href,
                               type,
                               url: href,
                             })
```

packages/article-skeleton/src/article-body/article-body-row.js
Geoff-Ford: Are you doing as part of this PR?

c-j-j: I think I'll do this in a subsequent PR... we already have inline styles for things like interactives/videos/etc, and these need to be checked thoroughly after refactoring on mobile/tablet/android/ios to check all looks good. We also need to pass `narrowContent` to the `styleFactory` to generate all styles... so it's quite a big change

packages/article-paragraph/src/index.js
c-j-j: Changing this to `!ast?.length` has caused snapshots to change - so I don't think that's the equivalent of `!ast || ast.length === 0`

c-j-j: Tests needed to be updated. Still weird that snapshots passed before.

Geoff-Ford (approved):

PR #344, opened 2021-02-05
fix: (TNLT-6511) scroll through images in order

kamiranoff:
- https://nidigitalsolutions.jira.com/browse/TNLT-6511
- No change in design but this allows user to swipe through images of an article in the order in the gallery for Magazine standard, main standard, and in-depth templates.

Let me know what screenshots you want me to add if any.

https://user-images.githubusercontent.com/5861100/107355332-34b14100-6ac7-11eb-9a9c-b59776f296f2.mov


aronshamash:
@kamiranoff could add a video of it ?

packages/utils/src/get-extra-images-content.ts
```diff
@@ -1,13 +1,54 @@
...
+  }
+
+  if (leadAsset.__typename === "Image") {
+    crop = getStandardTemplateCrop(leadAsset);
+  }
+
+  return crop;
 };
 
 export const getExtraImagesContent = (article: Article) => {
```
aronshamash, line 25: Can we find a better name for this function. I wasn't sure if this was get image for Article Extras, or get inline images or what?

packages/utils/src/get-lead-asset.js
```diff
@@ -7,16 +7,19 @@ export const defaultAsset = {
   leadAsset: null,
 };
 
+const getDisplayImage = (leadAsset, isVideo) => {
+  if (isVideo && leadAsset.posterImage) {
+    return getStandardTemplateCrop(leadAsset.posterImage);
```
aronshamash, line 12: Is this function getting the LeadAsset? Can we rename I was a bit unsure what it was doing ?

packages/utils/src/get-extra-images-content.ts
```diff
@@ -1,13 +1,54 @@
...
+  if (!crop) {
     return [];
   }
 
-  return article.content.filter((c) => c.name === "image");
+  return [
+    {
+      attributes: {
+        display: "",
+        caption: leadAsset.caption,
```
aronshamash, line 42: don't we need to do the same checks as for `credits` for `caption` field

packages/utils/src/is-template-with-lead-asset-in-gallery.ts
```diff
@@ -0,0 +1,6 @@
+import { Template } from "@times-components-native/article-skeleton/types";
+
+export const isTemplateWithLeadAssetInGallery = (template: Template) =>
+  template === Template.MAIN_STANDARD ||
```
aronshamash, line 4: what about having an array of supported templates and just use `includes` ? Less logic.

aronshamash: is this to know if the template supports having a lead asset in the gallery ?

packages/utils/src/get-extra-images-content.ts
kamiranoff: it gets all images of an article including the lead asset. `getImages` ? or `getAllImages`?

packages/utils/src/is-template-with-lead-asset-in-gallery.ts
kamiranoff: yes, that will be better. 
And yes. Some templates need Lead Asset in the gallery. The others don't have lead assets or only a small author picture.

packages/utils/src/get-extra-images-content.ts
kamiranoff: The extra check is made for credits because `credits` exists in the `Image` interface but not in the `Video` interface. While caption is in both hence no need for the `in` notation
TypeScript was complaining about `credits not existing in Type Media`.

But the case is handled for both if they are undefined/null

packages/utils/src/get-extra-images-content.ts
aronshamash: `getAllArticleImages` ?

packages/utils/src/is-template-with-lead-asset-in-gallery.ts
aronshamash: ok makes sense - they count those small roundel headshot as the lead asset yes I remember now. 👍

aronshamash (approved):

packages/article-skeleton/src/body-utils/index.js
```diff
@@ -3,12 +3,49 @@ import memoize from "memoize-one";
...
+  if (!skeletonProps?.data) {
+    return skeletonProps;
+  }
+
+  if (isTemplateWithLeadAssetInGallery(skeletonProps.data.template)) {
+    // the lead asset image will assume index 0
+    index = 1;
+  }
+
+  const content = skeletonProps.data.content.map((node) => {
```
c-j-j, line 22: Just throwing this out there, but if you wanted to avoid mutating an index variable in the `map` function - something like the following could also work. Up to you though

```
  if (!skeletonProps?.data) {
    return skeletonProps;
  }

  const offset = isTemplateWithLeadAssetInGallery(skeletonProps.data.template) ? 1 : 0;

  const content = skeletonProps.data.content.map((node, index) => {
    if (node.name !== "image" || node.attributes?.imageIndex) {
      return node;
    } else {
       return {
      attributes: {
        ...node.attributes,
        imageIndex: index + offset,
      },
    }
    }
  });
```

packages/fixture-generator/src/types.ts
```diff
@@ -2826,3 +2882,15 @@ export type StandardSectionSlice =
...
+  crop11?: Crop | null;
+  crop169?: Crop | null;
+  crop23?: Crop | null;
+  crop32?: Crop | null;
+  crop45?: Crop | null;
+  crop54?: Crop | null;
+  crop1251?: Crop | null;
+  id: string;
+  title: string;
+}
```
c-j-j, line 2896: just double checking, did you run the schema script to update this file - or did you manually change something?

packages/utils/src/crop-config.ts
```diff
@@ -0,0 +1,13 @@
+// MOVE THIS BACK INTO THE TEMPLATE FILE AS ITS SPECIFIC TO IT
```
c-j-j, line 1: Is this comment still needed?

packages/utils/src/crop-config.ts
```diff
@@ -0,0 +1,13 @@
...
+import {
+  TimesImage,
+  Video,
+} from "@times-components-native/fixture-generator/src/types";
+
+export default (crops: TimesImage | Video["posterImage"]) => {
+  if (!crops) {
+    return null;
+  }
+  const { crop169, crop32, crop1251, crop11, crop45, crop23 } = crops;
```
c-j-j, line 11: Should we export this function as a named export (as it's a new file?)?

packages/fixture-generator/src/types.ts
kamiranoff: Aj no I did it manually. Does it need to be generated?

packages/article-skeleton/src/body-utils/index.js
kamiranoff: ah, that'd be a lot better indeed.

packages/utils/src/crop-config.ts
kamiranoff: It is used in two different files within the utils folder. I guess it is not needed anymore.

packages/utils/src/crop-config.ts
kamiranoff: Good shout. it is used as `getStandardTemplateCrop` now, but I guess the name could be improved. `getCropByPriority`, `getPriorityCrop` ?

packages/fixture-generator/src/types.ts
kamiranoff: ah just read the readme. will amend

packages/utils/src/crop-config.ts
c-j-j: getCropByPriority sounds good to me

aronshamash: yes much better

packages/article-skeleton/src/body-utils/index.js
kamiranoff: hum. I remember. The reason I was not taking index from the loop is that it loops from all nodes.  I want to get only the index of images and only increment it for every images..

c-j-j: Ah right, good point. Yeah feel free to leave it as is then.

c-j-j (approved):

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "3.0.1",
+  "version": "3.0.1-beta",
```
kamiranoff, line 4: !! revert this before merging.

PR #347, opened 2021-02-09
chore: (TNLT-6479) refactor inline styles to stylesheet objects

c-j-j:
follow on from #343 

https://nidigitalsolutions.jira.com/browse/TNLT-6479

### DIFF
--- a/packages/article-skeleton/src/styles/article-body/shared.js
+++ b/packages/article-skeleton/src/styles/article-body/shared.js
@@ -1,7 +1,18 @@
 import styleguide, { tabletWidth } from "@times-components-native/styleguide";
 
-const sharedStyles = (scale) => {
+const sharedStyles = ({ scale, narrowContent, fontScale }) => {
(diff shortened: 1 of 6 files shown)


packages/article-skeleton/src/styles/article-body/shared.js
```diff
@@ -41,6 +52,11 @@ const sharedStyles = (scale) => {
       alignSelf: "center",
       width: tabletWidth,
     },
+    defaultFont: defaultFont,
```
Geoff-Ford, line 55: Is this because it styling or can it just be `defaultFont,`?

Geoff-Ford (approved):

kamiranoff (approved):

packages/article-skeleton/src/styles/article-body/shared.js
c-j-j: good spot - i'll change

PR #349, opened 2021-02-11
fix: (TNLT-6674) memoize article content

c-j-j:
Part of https://nidigitalsolutions.jira.com/browse/TNLT-6674

## Problem

Whenever we render the article AST, it appears that the entire component tree is being destroyed and recreated. This is causing issues like ad's re-rendering and showing a different image whilst the user is scrolling.

Also, whenever an article was marked as read, that state change was causing a re-render of the article.

## Solution

1. Extract the components that renders the article AST into its own component such that it's not impacted by the `hasBeenRead` state changes.
2. Memoise this new wrapped component to prevent unnecessary re-renders.

### DIFF
--- a/packages/article-skeleton/src/article-skeleton.js
+++ b/packages/article-skeleton/src/article-skeleton.js
@@ -17,12 +17,11 @@ import fixup from "./body-utils";
 import ErrorBoundary from "./boundary";
 import { useResponsiveContext } from "@times-components-native/responsive";
 
-const ArticleWithContent = (props) => {
+const MemoisedArticle = React.memo((props) => {
   const {
     Header,
     data,
     analyticsStream,
-    onArticleRead,
     onCommentGuidelinesPress,
     onCommentsPress,
     onTooltipPresented,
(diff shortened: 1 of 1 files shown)


kamiranoff (approved):

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -17,12 +17,11 @@ import fixup from "./body-utils";
 import ErrorBoundary from "./boundary";
 import { useResponsiveContext } from "@times-components-native/responsive";
 
-const ArticleWithContent = (props) => {
+const MemoisedArticle = React.memo((props) => {
```
kamiranoff, line 20: you don't want to create a new file for it?

c-j-j: Wasn't sure really... it seems very coupled with this article-skeleton component and takes in the same props, etc, so not sure how much extracting to a different file will help things?

kamiranoff: Yes, I guess it is primarily opinion-based according to:
is it more readable?
is it more testable?
is it more reusable?

c-j-j: True... I would say it's not reusable, and is tested in the context of article skeleton. It's really just extracted into a separate component as a performance optimisation

aronshamash (approved):

PR #350, opened 2021-02-11
chore: ad optimisations

c-j-j:
## Problem

We are seeing ads flicker on UA as you're scrolling past them. This is because the ad's webview is unnecessarily re-rendering, which sometimes loads in a brand new ad.

## Solution

1. Removes redundant AdComposer component
2. Will remove react-broadcast dependency that AdComposer required
3. Removed getDerivedStateFromProps from Ad component as was causing unnecessary re-renders
4. Removed height from DOMContext webview as was causing webview to refresh/change ad once it's loaded.

image

image


kamiranoff (approved):
Not familiar with this ad thing. but LGTM.

kamiranoff:
@c-j-j lint check has failed.

packages/ad/src/dom-context.js
```diff
@@ -148,7 +148,7 @@ class DOMContext extends PureComponent {
     const html = `
       <html>
         <head>
-        <meta name="viewport" content="width=${width},height=${height},initial-scale=1,user-scalable=no">
```
aronshamash, line 151: don't we need the `width` ?

c-j-j: It seems to work without width/height @aronshamash  - I tested on iOS/android/tablet/mobile and all seems to work. I'm really unclear why this was there

aronshamash: in generally it can improve perf because the renderer knows dimensions upfront but I guess if it didn't make a difference lets leave as it is.

c-j-j: `width` may be okay to leave in, but `height` definitely causes a re-render which effectively refreshes the webview

c-j-j: and if it works without then would be good to leave both out

aronshamash (approved):

PR #352, opened 2021-02-12
fix: (TNLT-6708) calculates section content height with RN

c-j-j:
https://nidigitalsolutions.jira.com/browse/TNLT-6708

- [x] iOS repo needs new dependency added (https://github.com/newsuk/nuk-tnl-app-ios-universal/pull/2327)

## Problem

We had performance issues having iOS calculate the section content height. We have changed the implementation to approximately calculate within RN.

### iPad 8th Gen (no homebar)
image
image

### iPad Air 4th Gen (with homebar)

simulator_screenshot_53234531-905D-462F-9C68-0CE05AA68120
simulator_screenshot_2B7AE0AE-EEF2-429E-B937-62081EC20A83


packages/responsive/__tests__/shared.base.js
```diff
@@ -26,4 +27,16 @@ export default () => {
 
     expect(testInstance).toMatchSnapshot();
   });
+
+  it("calculates section content height", () => {
+    const { sectionContentHeight } = calculateResponsiveContext(500, 1000, 1);
+
+    //  1000 (screen height)
+    // -  10 (bottom inset)
+    // -  5  (top inset)
```
Geoff-Ford, line 36: Your `5` isn't lined up correctly :trollface:

packages/responsive/src/calculateResponsiveContext.ts
```diff
@@ -4,28 +4,37 @@ import {
...
-  isLandscape: width > height,
-  sectionContentWidth: contentSize?.width ?? 0, // todo remove `?? 0` once android also passes contentSize
-  sectionContentHeight: contentSize?.height ?? 0,
-});
+): ContextType => {
+  const contentHeight =
+    height -
+    ((initialWindowMetrics?.insets.bottom ?? 0) +
+      (initialWindowMetrics?.insets.top ?? 0)) -
+    approximateNavHeightOnTablet;
```
Geoff-Ford, line 23: Is this not tablet specific?

package.json
```diff
@@ -94,6 +94,7 @@
     "react-native-device-info": "7.3.1",
     "react-native-image-zoom-viewer": "3.0.1",
     "react-native-navigation": "^6.11.0",
+    "react-native-safe-area-context": "^3.1.9",
```
c-j-j, line 97: need to hardcode this to 3.1.9

packages/responsive/src/calculateResponsiveContext.ts
c-j-j: This is the nav height on tablet

Geoff-Ford: I updated the comment above

c-j-j: Yeah good point, i can rename `sectionContentHeight` to `sectionContentHeightTablet` to very clearly show this is tablet specific?

package.json
Geoff-Ford: Yeah, I see that latest "release is `3.1.8` but the latest tag is `3.1.9`. Any idea why this wasn't yet release? Guessing you also need to latest tag to for your fix to work correctly?

packages/responsive/src/calculateResponsiveContext.ts
```diff
@@ -4,28 +4,37 @@ import {
...
+    fontScale,
+    isTablet:
+      (config && config.breakpoint && config.breakpoint !== "small") ||
+      width >= minTabletWidth,
+    windowWidth: width,
+    windowHeight: height,
+    orientation: height > width ? Orientation.PORTRAIT : Orientation.LANDSCAPE,
+    isPortrait: height > width,
+    isLandscape: width > height,
+    sectionContentWidth: width,
```
Geoff-Ford, line 37: So no impact on using just `width` here rather than the previous `contentSize?.width ?? 0`?

package.json
c-j-j: Not sure, this is the version after doing `yarn add react-native-safe-area-context` - and is latest according to https://www.npmjs.com/package/react-native-safe-area-context - maybe their github releases aren't updated?

packages/responsive/src/calculateResponsiveContext.ts
Geoff-Ford: or maybe just fall back to `height` if not a tablet? I see you are using the `width` now for `sectionContentWidth`

packages/responsive/src/calculateResponsiveContext.ts
c-j-j: Seems to work fine (see screenshots for portrait which uses sectionContentWidth)

packages/test-utils/src/responsiveContextUtil.tsx
```diff
@@ -3,12 +3,7 @@ import { calculateResponsiveContext } from "@times-components-native/responsive/
 import React from "react";
 
 const withResponsiveContext = (WrappedComponent: any, width: number) => (
-  <ResponsiveContext.Provider
-    value={calculateResponsiveContext(width, 400, 1, {
-      width: 1024,
-      height: 600,
-    })}
-  >
+  <ResponsiveContext.Provider value={calculateResponsiveContext(width, 400, 1)}>
```
kamiranoff, line 6: Minor point. Might be worth add `400` and `1` to variables/constants to make sure we know what they are referring to

package.json
Geoff-Ford: 🤷‍♂️

packages/responsive/src/calculateResponsiveContext.ts
c-j-j: `height` isn't really correct for tablet or mobile though. `width` just happens to be correct as a section spans the full screen width.

c-j-j: Maybe I can leave as is, and leave a comment that this is for tablet only? We might revisit and come up with a better approach that works for tablet/mobile in future. `sectionContentHeightTablet` is quite clunky otherwise

aronshamash: would go with explicit var name rather than a comment

aronshamash: anyone digging into the code will end up at this PR for ref.

c-j-j: Ok. So will go with `sectionContentWidth` and `sectionContentHeightTablet` to be specific

kamiranoff (approved):

Geoff-Ford (approved):

PR #355, opened 2021-02-15
feat: (TNLT-6397) Article read animation

aronshamash:
- https://nidigitalsolutions.jira.com/browse/TNLT-6397
- For articles that have been deemed as being read we need to differentiate the article tiles that have already be shown to the user (`highlight:false`) and recently read articles which have not yet been shown (`highlight:true`). 
- The `highlight` flag allows TCN to distinguish which article tiles should fade out to reduced opacity and which should have reduced opacity initially without animation.
- `Tile AH` used by the Cartoon slice needed some refactoring because the tile was rendering the byline using the `Byline` component rather than `ArticleSummary` like the rest of the other tiles.
- `isOpinionByline` flag is not used so it was removed from `ArticleSummary`

Agreed data format:

```
    const readArticles = [
      {
        highlight: true,
      },
      {
        highlight: false,
      },
    ];
```

### iOS
https://user-images.githubusercontent.com/1736782/108087978-dd374600-706f-11eb-8e4d-0b4bcfba3c44.mov

### Android

https://user-images.githubusercontent.com/1736782/108191323-e4f7f880-710a-11eb-8286-a4a9c73d5aed.mov


packages/pages/src/section/section.js
```diff
@@ -142,7 +142,24 @@ class SectionPage extends Component {
       sectionName: section.name,
     });
 
-    const readArticles = [];
+    const readArticles = [
```
aronshamash, line 145: remove before merging

packages/edition-slices/src/tiles/shared/tile-summary.tsx
```diff
@@ -64,158 +56,37 @@ const TileSummary: React.FC<Props> = ({
...
-                  markAsRead,
-                }}
-                strapline={renderStrapline(markAsRead)}
-                saveStar={withStar && renderSaveStar()}
-                style={style}
-              />
-            );
-          }}
+          {({ readArticles }) => (
+            <ArticleSummary
```
aronshamash, line 64: Now have a component in order to pass in `readArticles` as a prop so don't need to worry about managing state.

packages/article-summary/src/article-summary.js
```diff
@@ -86,13 +98,26 @@ function ArticleSummary(props) {
...
     <View style={style}>
-      {labelProps ? <ArticleSummaryLabel {...labelProps} /> : null}
-      {isOpinionByline && byline}
+      {labelProps ? (
+        <ArticleSummaryLabel
+          {...labelProps}
+          articleReadState={articleReadState}
+        />
+      ) : null}
+      {bylineOnTop && byline}
```
kamiranoff, line 120: I think you have the opposite behaviour in line 131?
```
{!bylineOnTop && byline}
```

kamiranoff: Oh. Ok. I understand.

packages/article-summary/src/article-summary.js
```diff
@@ -21,18 +20,42 @@ function renderAst(ast) {
...
       delay: ARTICLE_READ_ANIMATION.delay,
       duration: ARTICLE_READ_ANIMATION.duration,
-      toValue: 0.6,
+      toValue: opacity,
       useNativeDriver: true,
     }).start();
-  }, [markAsRead]);
+  }, [articleReadState]);
+
+  return articleReadState.animate ? (
```
kamiranoff, line 36: Minor but might be clearer to have two if/ switch statement rather than a nested ternary.

kamiranoff (approved):

c-j-j (approved):

packages/article-summary/src/article-summary.js
aronshamash: I think I'll leave as is this time. It seems to read ok to other devs.

packages/pages/src/section/section.js
aronshamash: done

PR #357, opened 2021-02-17
feat: Context retention

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-6405
https://nidigitalsolutions.jira.com/browse/TNLT-6436

Takes an article ID provided by native in a section's initProps, searches for it in all slices of the section, and if found returns the index of the slice and scrolls to that slice.

Android:
https://user-images.githubusercontent.com/5103512/108190641-1e7c3400-710a-11eb-8e2b-c266c51219aa.mov

iOS:
https://user-images.githubusercontent.com/5103512/108192995-d90d3600-710c-11eb-8989-f707d2b9d3b4.mov


packages/section/src/utils/addToFlatlistOffset.ts
```diff
@@ -0,0 +1,23 @@
...
+  sliceOffsets[sliceIndex] = height;
+  if (Object.keys(sliceOffsets).length === sliceOffsetIndex) {
+    const sliceOffset = Object.values(sliceOffsets).reduce(
+      (acc: number, height: number) => acc + height,
+      0,
+    );
+    setTimeout(() => {
+      flatListRef.current?.scrollToOffset({ offset: sliceOffset });
+      sliceOffsets = {};
+    }, 1000);
```
Geoff-Ford, line 21: This can be whatever we want. Let's see what MM thinks.

packages/section/src/section.tsx
```diff
@@ -153,9 +172,22 @@ const Section: React.FC<Props> = ({
 
   if (slices) receiveChildList(data);
 
+  // scrollToArticleId =
+  //   sectionTitle === "News"
+  //     ? "eb796c0a-6f9f-11eb-811f-f64a7b4cb430"
+  //     : undefined;
+  // News = "eb796c0a-6f9f-11eb-811f-f64a7b4cb430"
+  // Times2 = "075c38fa-6fb2-11eb-bd2f-f33f509764cd"
+  // scrollToArticleId = "cd1cd32c-6ef6-11eb-bd2f-f33f509764cd";
```
Geoff-Ford, line 181: I'll remove this commented override once I grab another screen vid

packages/pages/src/section/section.js
```diff
@@ -135,7 +135,12 @@ class SectionPage extends Component {
   }
 
   render() {
-    const { publicationName, publishedTime, variants } = this.props;
+    const {
+      publicationName,
+      publishedTime,
+      scrollToArticleId,
```
kamiranoff, line 141: I think the naming implies a function. If it is a string maybe `IdToScrollTo`?

packages/section/src/section.tsx
```diff
@@ -169,7 +193,7 @@ const Section: React.FC<Props> = ({
         nestedScrollEnabled
         onViewableItemsChanged={onViewed ? onViewableItemsChanged : null}
         {...(isPuzzle && { onScrollBeginDrag })}
-        renderItem={renderItem(isPuzzle, orientation)}
+        renderItem={renderItem(isPuzzle, orientation, sliceIndexFromArticle)}
```
kamiranoff, line 196: wouldn't be possible to simplify with `renderItem={renderItem}` and remove the extra function all in `const renderItem`?

packages/section/src/section.tsx
```diff
@@ -90,24 +96,36 @@ const Section: React.FC<Props> = ({
...
-      adConfig={adConfig}
-      sectionTitle={sectionTitle}
-      orientation={orientation}
-      isTablet={isTablet}
-    />
+  const renderItem = (
+    isPuzzle: boolean,
+    orientation: Orientation,
+    sliceOffsetIndex: number,
+  ) => ({ index, item: slice, inTodaysEditionSlice }: any) => (
```
kamiranoff, line 103: I was thinking you change this to
```
const renderItem = (
    isPuzzle: boolean,
    orientation: Orientation,
    sliceOffsetIndex: number,
  ): ListRenderItem<any> => ({ index, item: slice, inTodaysEditionSlice }) => (
  ```
  And maybe change `any` if we have the Item Type defined but TypeScript still complains about `inTodaysEditionSlice`. Is that defined somewhere?

File: packages/pages/src/section/section.js (hunk 3 above)
Comment (Geoff-Ford) [reply packages/pages/src/section/section.js:141]: Take it up with @aronshamash , he named the baby :trollface:

File: packages/section/src/section.tsx (hunk 4 above)
Comment (Geoff-Ford) [reply packages/section/src/section.tsx:196]: It is curried.

File: packages/section/src/section.tsx (hunk 5 above)
Comment (Geoff-Ford) [reply packages/section/src/section.tsx:103]: Yeah, we don't currently have a `slice` type as that would be pretty complex as so many variations. Needs doing at some point though.

File: packages/section/src/section.tsx (hunk 4 above)
Comment (c-j-j) [reply packages/section/src/section.tsx:196]: Appreciate this is adding onto what's there, but I'm not sure why this is curried to begin with as all 3 args are in the same scope as renderItem.

File: packages/section/src/section.tsx [hunk 6]
```diff
@@ -140,6 +158,7 @@ const Section: React.FC<Props> = ({
     return renderItem(
       false,
       orientation,
+      0,
```
c-j-j, line 161: You could set this to `sliceIndexFromArticle` (after moving that block up above this) to remove the magic number here.

packages/section/src/section.tsx
kamiranoff: I could be wrong but it seems to me that `inTodaysEditionSlice ` is always undefined inside of here?

packages/section/src/utils/addToFlatlistOffset.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export let sliceOffsets: Record<number, number> = {};
+
+export const addToFlatlistOffset = (
+  sliceIndex: number,
+  height: number,
+  sliceOffsetIndex: number,
+  flatListRef: RefObject<FlatList>,
+) => {
+  sliceOffsets[sliceIndex] = height;
```
c-j-j, line 12: This kind of data would normally be stored in react state/context. What's the reason for using a global variable for this?

Geoff-Ford: This is a self containing helper method, not a component. Also, closure does what is expected here without the need for the method to return its findings to then be put in state. Having said that, this is about to change.

packages/section/src/utils/addToFlatlistOffset.ts
aronshamash: worth putting in a const when we have agreed on the timeout

packages/section/src/section.tsx
```diff
@@ -95,19 +119,28 @@ const Section: React.FC<Props> = ({
...
-      inTodaysEditionSlice={inTodaysEditionSlice}
-      adConfig={adConfig}
-      sectionTitle={sectionTitle}
-      orientation={orientation}
-      isTablet={isTablet}
-    />
+    <View
+      onLayout={(event) => {
+        const height = event?.nativeEvent?.layout?.height ?? 0;
+        sliceOffsets.current[index] = height;
```
aronshamash, line 125: can do this on just 1 line no?

packages/section/src/section.tsx
```diff
@@ -95,19 +119,28 @@ const Section: React.FC<Props> = ({
...
+      }}
+      style={{ flex: 1 }}
+      testID="sliceRenderView"
```
aronshamash, line 128: I take it this doesn't need to be unique?

packages/section/src/section.tsx
```diff
@@ -153,9 +186,29 @@ const Section: React.FC<Props> = ({
 
   if (slices) receiveChildList(data);
 
+  const scrollToOffset = (event: { articleId: string }) => {
+    const { articleId } = event;
+    const sliceIndexFromArticle = articleId
```
aronshamash, line 191: can do the check for `articleId` in the `getSliceIndexByArticleId ` util ?

aronshamash (approved):
looks good to me.

packages/section/src/section.tsx
Geoff-Ford: I could but it would wrap the line anyway so I went with clarity. Can change though?

packages/section/src/section.tsx
Geoff-Ford: Actually that isn't needed anymore at all so good spot 😉 . I'll remove.

packages/section/src/section.tsx
Geoff-Ford: You mean check if `articleId` has been passed in to `getSliceIndexByArticleId`? If it isn't, it just won't be found (`articleId === item.articleId`).

packages/section/src/section.tsx
aronshamash: nah is fine.

Geoff-Ford: Actually that does fit so have changed it.

packages/section/src/section.tsx
aronshamash: yes I'm thinking we can avoid having to check for articleId and returning 0 here

Geoff-Ford: My thinking was that if I know it isn't there, no need to call the function.

aronshamash: well this should never happen in reality and is handled in `getSliceIndexByArticleId` so personally would just remove it but not a biggie.

packages/provider-queries/src/section_fragment.graphql
```diff
@@ -334,6 +334,9 @@ fragment commonSlicesSection on StandardSectionSlice {
     }
   }
   ... on LeadOneAndFourSlice {
+    items {
```
c-j-j, line 337: LeadersSlice/CommentLeadAndCartoonSlice /StandardSlice don't have this added - won't they miss out on this context retention? If possible, it would be nice to have this extracted into the common fields of a slice (not necessarily for this PR)?

packages/section/src/section.tsx
```diff
@@ -95,19 +119,26 @@ const Section: React.FC<Props> = ({
...
-      adConfig={adConfig}
-      sectionTitle={sectionTitle}
-      orientation={orientation}
-      isTablet={isTablet}
-    />
+    <View
+      onLayout={(event) => {
+        sliceOffsets.current[index] = event?.nativeEvent?.layout?.height ?? 0;
+      }}
+      style={{ flex: 1 }}
```
c-j-j, line 126: Inline style probably should be extracted into a stylesheet object.

c-j-j (approved):
Nicely done - approach looks sensible to me 👍

packages/section/src/section.tsx
kamiranoff, line 126: Not a biggie but a bit of online style.

kamiranoff (approved):

packages/section/src/section.tsx
Geoff-Ford: For just this? Man you are brutal

packages/provider-queries/src/section_fragment.graphql
Geoff-Ford: That is a good question. I took the query changes from @aronshamash 's POC.

packages/section/src/section.tsx
c-j-j: I thought it helped a bit with performance. I did say _probably_

kamiranoff: I think so too, has it doesn't recreate a new object for every render. I concord with the probably.

Geoff-Ford: Brutal, the pair of you, at this time on a Friday!

Geoff-Ford: Done 😉

packages/provider-queries/src/section_fragment.graphql
aronshamash: think standard slice is used for Related articles on article page btw.

PR #358, opened 2021-02-18
feat: Integration of article read with all templates

aronshamash:
- Integration of `onArticleRead` bridge call for all article templates.


packages/pages/src/article/article-base.js
```diff
@@ -71,9 +72,7 @@ const ArticleBase = ({
           error={omitErrors ? null : error}
           interactiveConfig={interactiveConfig}
           isLoading={isLoading || (omitErrors && error)}
-          onArticleRead={() => {
-            // @TODO onArticleRead
-          }}
+          onArticleRead={(id) => onArticleRead(id)}
```
c-j-j, line 75: Could also be written as `onArticleRead={onArticleRead}` - up to you

c-j-j (approved):

kamiranoff (approved):

packages/pages/src/article/article-base.js
aronshamash: yes that is better will update. I had it with 1st param as `event` that why it is like this but yes not needed now

Geoff-Ford (approved):

PR #361, opened 2021-02-22
feat: Change article read data to event based

aronshamash:
- Current implementation of `readArticle` is via initial props but this will not work with the new Context Retention feature so agreed implementation should be event based.
- This is a BREAKING CHANGE
- a new `updateReadArticles` event will be fired by native layer on section pages. The payload will be an array in the following format:

```
[
  {
    highlight: true,
  },
  {
    highlight: false,
  },
]
   ```

### DIFF
--- a/packages/pages/src/section/section.js
+++ b/packages/pages/src/section/section.js
@@ -38,7 +38,7 @@ class SectionPage extends Component {
     const { section } = this.props;
     this.state = {
       recentlyOpenedPuzzleCount: props ? props.recentlyOpenedPuzzleCount : 0,
-      readArticles: props ? props.readArticles : [],
+      readArticles: [],
       savedArticles: null,
       section,
     };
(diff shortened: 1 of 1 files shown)


c-j-j (approved):

kamiranoff (approved):

Geoff-Ford (approved):
LGTM. Curious about the re-renders on section page and the section component and if this adversely affect the context retention (although it shouldn't)

aronshamash:
> LGTM. Curious about the re-renders on section page and the section component and if this adversely affect the context retention (although it shouldn't)

Yes it does force a re-render and in _some_ cases causes the context retention to not scroll correctly. I have created a new PR with different approach here: https://github.com/newsuk/times-components-native/pull/363

aronshamash:
Need a different approach to stop section component being re-rendered so closing PR.

PR #363, opened 2021-02-23
chore: (BREAKING CHANGE) Update read articles to be event based

aronshamash:
- Current implementation of `readArticle` is via initial props but this will not work with the new Context Retention feature so agreed implementation should be event based.
- a new `updateReadArticles` event will be fired by native layer on section pages but only used within the `ArticleSummary` component in `edition-slices`. The payload will be an array in the following format:

```
[
  {
    highlight: true,
  },
  {
    highlight: false,
  },
]
   ```
 - In order to know when an article is in focus we are now listening to a new event fired on the article page `onArticleFocus` which sends the articleId as the payload, allowing RN layer to check if the article id matches to know to start the article read timer.

### CODE REVIEW

**Review by kamiranoff — APPROVED** [review]

File: packages/article-skeleton/src/article-skeleton.js [hunk 1]
```diff
@@ -121,23 +129,38 @@ const MemoisedArticle = React.memo((props) => {
 
 const ArticleWithContent = (props) => {
   const { onArticleRead, data } = props;
+  const articleReadTimerDuration = 6000;
+  let hasBeenRead = false;
+  let articleReadDelay = null;
+
+  const setArticleReadTimeout = (articleId) => {
+    if (articleId === data.id && !hasBeenRead) {
+      articleReadDelay = setTimeout(() => {
```
kamiranoff, line 138: nitpicking but you could simpify with `setTimeout(setArticleRead, articleReadTimerDuration)` i believe.

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -121,23 +129,38 @@ const MemoisedArticle = React.memo((props) => {
...
+    return () => {
+      updateReadArticlesEventsListener.remove();
+    };
+  }, []);
 
   const setArticleRead = () => {
-    setHasBeenRead(true);
+    if (hasBeenRead) return;
+    hasBeenRead = true;
+    console.log("SET ARTICLE READ", data.url);
```
kamiranoff, line 160: console.log

packages/article-skeleton/src/article-skeleton.js
kamiranoff: Out of curiosity, why you need the event listener here? Couldn't you just called the timeout in the useEffect "onMount" below ?

packages/edition-slices/src/tiles/shared/article-summary.tsx
```diff
@@ -146,18 +156,27 @@ const ArticleSummary: React.FC<Props> = ({
...
+    const updateReadArticlesEventsListener = sectionEventEmitter.addListener(
+      "updateReadArticles",
+      updateReadArticles,
+    );
+    return () => {
+      updateReadArticlesEventsListener.remove();
+    };
+  }, []);
+
+  const updateReadArticles = (readArticles: [ArticleRead]) =>
```
kamiranoff, line 178: `[ArticleRead]` this notation is for Tuples. I think if you have an array of articles you want to use `ArticleRead[]`

packages/article-skeleton/src/article-skeleton.js
kamiranoff: Ah is it because the event is triggered natively? I don't see where you emit the `onArticleFocus ` event for example.

aronshamash: yes it relies on native to emit the `onArticleFocus` event.

aronshamash: this is because we have 5 articles rendered at any one time but only 1 displayed to the user - the rest are cached and ready. Therefore we didn't know which was the currently in focus article before native added this event.

packages/edition-slices/src/tiles/shared/article-summary.tsx
aronshamash: yes thanks : )

packages/article-skeleton/src/article-skeleton.js
kamiranoff: ah yes makes sense!

packages/article-skeleton/src/article-skeleton.js
```diff
@@ -121,23 +129,38 @@ const MemoisedArticle = React.memo((props) => {
...
+  let articleReadDelay = null;
+
+  const setArticleReadTimeout = (articleId) => {
+    if (articleId === data.id && !hasBeenRead) {
+      articleReadDelay = setTimeout(() => {
+        setArticleRead();
+      }, articleReadTimerDuration);
+      return () => {
+        clearTimeout(articleReadDelay);
+      };
```
Geoff-Ford, line 143: Can be on one line?
`return () => clearTimeout(articleReadDelay);`

Actually, why is this function being returned?

aronshamash: yes no good reason. will update.

packages/edition-slices/src/tiles/shared/article-summary.tsx
```diff
@@ -63,12 +68,22 @@ interface Props {
...
   hideLabel?: boolean;
   whiteSpaceHeight?: number;
 }
 
+export const getArticleReadState = (
+  isTablet: boolean,
+  readArticles: Array<ArticleRead> | null,
+  articleId: string,
+): ArticleReadState => ({
+  read: isTablet && !!readArticles?.some((obj) => obj.id === articleId),
```
Geoff-Ford, line 81: `.some` returns a boolean so guessing you are coercing to boolean because `readArticles` can be `null`? You can instead use the `null` coalesce so:

`read: isTablet && (readArticles?.some((obj) => obj.id === articleId) ?? false),`

Not convinced that is any more readable now I've written it, but seemed like the "modern" way to handle `null` or `undefined` values in optional chaining.

packages/edition-slices/src/tiles/shared/article-summary.tsx
```diff
@@ -63,12 +68,22 @@ interface Props {
...
+  animate:
+    isTablet &&
+    !!readArticles?.some((obj) => obj.highlight && obj.id === articleId),
```
Geoff-Ford, line 84: Same point as above if you are convinced

packages/styleguide/src/articleRead/index.ts
```diff
@@ -1,4 +1,4 @@
 export const ARTICLE_READ_ANIMATION = {
-  DELAY: 500,
-  DURATION: 500,
+  DELAY: 4000,
+  DURATION: 2000,
```
Geoff-Ford, line 3: Are these the actual new values?

packages/edition-slices/src/tiles/shared/article-summary.tsx
aronshamash: yes this looks better with null coalesce

packages/styleguide/src/articleRead/index.ts
aronshamash: I've updated now to increase delay and duration but by at lesser amount.

packages/article-skeleton/src/article-skeleton.js
aronshamash: removed

Geoff-Ford (approved):

PR #367, opened 2021-02-25
fix: (TNLT-6824): remove tracking section on ios

kamiranoff:
- https://nidigitalsolutions.jira.com/browse/TNLT-6824

- This removes tracking of sections from the react-native side as it will be handled natively in ios.

### DIFF
--- a/packages/pages/src/section/track-section.js
+++ b/packages/pages/src/section/track-section.js
@@ -1,6 +1,5 @@
 import { NativeModules } from "react-native";
 
-const { track } = NativeModules.ReactAnalytics;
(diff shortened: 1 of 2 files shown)


packages/pages/src/section/track-section.js
```diff
@@ -1,6 +1,5 @@
 import { NativeModules } from "react-native";
```
kamiranoff, line 1: this is only for ios as there is a `.android` file for android

aronshamash (approved):

Geoff-Ford (approved):

PR #369, opened 2021-03-01
fix: storybook app from crashing

kamiranoff:
- Storybook app was crashing on startup due to missing native modules.

### DIFF
--- a/ios/Native Modules/SectionEvents.m
+++ b/ios/Native Modules/SectionEvents.m
@@ -22,7 +22,47 @@ RCT_EXPORT_MODULE();
 
 RCT_EXPORT_METHOD(onSectionLoaded:(NSString *) sectionName event:(NSDictionary *) event)
 {
-  NSLog(@"onSectionLoaded  %@ %@", sectionName, event);
+  NSLog(@"onSectionLoaded %@ %@", sectionName, event);
+}
+
+RCT_EXPORT_METHOD(onArticlePress:(NSString *) url isPuff:(NSString *) isPuff)
+{
+  NSLog(@"onArticlePress %@ %@", url, isPuff);
+}
+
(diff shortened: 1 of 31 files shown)


Geoff-Ford:
Is there anything required for Android also?

kamiranoff:
> Is there anything required for Android also?

Ah, Good point. The app still running fine on Android. But will double-check if some components are crashing.

ios/Native Modules/SectionEvents.m
```diff
@@ -0,0 +1,28 @@
...
+//  TimesComponents
+//
+//  Created by Kevin Amiranoff on 01/03/2021.
+//
+
+#import <Foundation/Foundation.h>
+#import <React/RCTBridgeModule.h>
+#import <React/RCTEventEmitter.h>
+
+@interface SectionEvents : RCTEventEmitter <RCTBridgeModule>
```
kamiranoff, line 12: I think there are more events to be implemented in this class, but at least it is not crashing.

Geoff-Ford (approved):

ios/Native Modules/SectionEvents.m
aronshamash: so seems some events crash it others don't

aronshamash (approved):

ios/Native Modules/SectionEvents.m
kamiranoff: Ah I think it wasn't running at all because the module didn't exist. There still might be crashes for specific events if they are not mocked.

kamiranoff: I guess I could implement a few more here. Might still miss some events but I'll try to catch what I can

kamiranoff: I implemented the remaining methods. All the ones that were already implemented in android

PR #371, opened 2021-03-02
fix: (TNLT-6840) Ads not displaying in section screens on Android

jonnyandrew:
##### Problem

Adverts are not displaying on sections on Android

##### Solution

Viewport Aware View that helps to lazy load the ads needs to be contained by a `Viewport.Tracker` to work. Without that, `onViewportEnter` never fires and ads are never loaded.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "5.0.2-beta0",
+  "version": "5.0.2-beta",
   "description": "A collection of react-native components for The Times and Sunday Times",
   "main": "index.js",
   "engines": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "4.0.0",
+  "version": "4.0.1-beta0",
   "description": "A collection of react-native components for The Times and Sunday Times",
   "main": "index.js",
   "engines": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "5.0.2-beta",
+  "version": "5.0.1",
   "description": "A collection of react-native components for The Times and Sunday Times",
   "main": "index.js",
   "engines": {
--- a/packages/section/src/section.tsx
+++ b/packages/section/src/section.tsx
@@ -28,7 +28,7 @@ import {
 import { OnArticlePress } from "@times-components-native/types";
 import { SectionTitles } from "./utils/sectionConfigs";
 import { Orientation } from "@times-components-native/responsive/src/context";
-import { Viewport } from "@skele/components"
+import { Viewport } from "@skele/components";
 
 const styles = styleFactory();
 const { SectionEvents } = NativeModules;
--- a/packages/section/src/section.tsx
+++ b/packages/section/src/section.tsx
@@ -28,6 +28,7 @@ import {
 import { OnArticlePress } from "@times-components-native/types";
 import { SectionTitles } from "./utils/sectionConfigs";
 import { Orientation } from "@times-components-native/responsive/src/context";
+// @ts-ignore
 import { Viewport } from "@skele/components";
 
 const styles = styleFactory();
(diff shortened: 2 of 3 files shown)


packages/section/src/section.tsx
```diff
@@ -206,23 +208,25 @@ const Section: FC<Props> = ({
...
-          renderItemSeperator(isPuzzle)(leadingItem, editionBreakpoint)
-        }
-        keyExtractor={(item) => item.elementId}
-        ListHeaderComponent={getHeaderComponent(isPuzzle, isMagazine)}
-        nestedScrollEnabled
-        onViewableItemsChanged={onViewed ? onViewableItemsChanged : null}
-        {...(isPuzzle && { onScrollBeginDrag })}
-        renderItem={renderItem(isPuzzle, orientation)}
-      />
+      <Viewport.Tracker>
```
Geoff-Ford, line 211: Great find. Curious why this worked on iOS without this though.

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "5.0.1",
+  "version": "5.0.2-beta0",
```
kamiranoff, line 4: Ah smart! :)

packages/section/src/section.tsx
```diff
@@ -28,6 +28,8 @@ import {
 import { OnArticlePress } from "@times-components-native/types";
 import { SectionTitles } from "./utils/sectionConfigs";
 import { Orientation } from "@times-components-native/responsive/src/context";
+// @ts-ignore
```
kamiranoff, line 31: Do you know what exact TypeScript rule is complaining about?. Might be better fixing the error or, ignoring the specific rule rather than the entire TypeScript linting.

packages/section/src/section.tsx
jonnyandrew: In `dom-context.js` the viewport aware view is used to set a `loaded` flag and when that flag is set, the advert will load. But on iOS there is an override for that behaviour that looks like

```
 {(Platform.OS === "ios" || loaded) && (
     // WebView
 ```
 
So it means the viewport aware view doesn't need to work properly on iOS for ads to work. But I guess that also means that a lot more ads are loading that is necessary for iOS users?

package.json
Geoff-Ford, line 4: Don't forget to remove this once tested

packages/section/src/section.tsx
Geoff-Ford: Ha, I had a thought about the same check and shared that with you on Slack lol.

packages/section/src/section.tsx
jonnyandrew: I am not totally sure to be honest - this is the error: https://app.circleci.com/pipelines/github/newsuk/times-components-native/2096/workflows/ce4c4214-dd92-4550-9482-47bc01ae6c12/jobs/11959/parallel-runs/0/steps/0-103?invite=true

I have copied the `@ts-ignore` solution from `packages/tooltip/tooltip.tsx` but perhaps there is a better way

package.json
Geoff-Ford: Okay, so as you have proved it works locally on Android, and I can confirm this branch works on iOS locally too, let's remove the beta change as not needed.

jonnyandrew: Thank you - I have reverted the version change: 019bb0a

packages/section/src/section.tsx
kamiranoff: Ah yes. Typescript needs a Type declaration file with the types of public methods/constants etc.. Some libraries come with a declaration file, some needs to be installed independently from this package. `https://github.com/DefinitelyTyped/DefinitelyTyped` some are in javascript and don't have a declaration file at all. 

TypeScript is complaining that he doesn't have the types for this library/import

kamiranoff: The library is in js and I can't find any type for it, so I guess it is fine.
more on this: https://blog.atomist.com/declaration-file-fix/

jonnyandrew: Thanks, well I have learned something today about JS and TS! I guess since this library is still in alpha, they might add type declarations in future

Geoff-Ford: I doubt they will. To be honest, our dependency on this library has always made me nervous since I found out it hasn't been touched in 2 years! Probably worth looking for a suitable replacement at some point.

https://www.npmjs.com/package/@skele/components

Geoff-Ford (approved):

kamiranoff (approved):

packages/section/src/section.tsx
aronshamash: btw I did look into this and other options and it is really thin on the ground. This one seems to be the most used out there.

PR #374, opened 2021-03-03
fix: Context retention on section page

aronshamash:
- fix context retention on section page

### DIFF
--- a/packages/section/src/section.tsx
+++ b/packages/section/src/section.tsx
@@ -67,7 +67,7 @@ const Section: FC<Props> = ({
   const { cover, name, slices, title: sectionTitle } = section;
   const { isTablet, editionBreakpoint, orientation } = useResponsiveContext();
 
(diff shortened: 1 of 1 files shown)


packages/section/src/section.tsx
```diff
@@ -210,7 +210,7 @@ const Section: FC<Props> = ({
     <>
       <Viewport.Tracker>
         <FlatList
-          ref={flatListRef}
+          ref={(ref) => (flatListRef.current = ref)}
```
kamiranoff, line 213: isn't it the old way of setting the ref? Sorry but, what is the issue?

Geoff-Ford: I saw the same fix. The issue is with the `Viewport` library preventing the `ref` being read. This is a suitable workaround.

kamiranoff (approved):

Geoff-Ford (approved):

PR #378, opened 2021-03-09
fix: (TNLT-6970) Fixes article and section judder when returning app to foreground

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-6970

### DIFF
--- a/packages/responsive/src/responsive.tsx
+++ b/packages/responsive/src/responsive.tsx
@@ -1,10 +1,10 @@
-import React, { useContext, useEffect, useState } from "react";
+import React, { useContext, useEffect, useRef, useState } from "react";
 import {
   getDimensions,
   addDimensionsListener,
   removeDimensionsListener,
 } from "@times-components-native/utils";
-import { ScaledSize } from "react-native";
+import { AppState, ScaledSize } from "react-native";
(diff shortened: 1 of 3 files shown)


packages/responsive/src/responsive.tsx
```diff
@@ -15,19 +15,32 @@ interface DimensionChangeEvent {
...
   const [state, setState] = useState(
     calculateResponsiveContext(width, height, fontScale),
   );
 
   const onDimensionChange = ({
     window: { fontScale, width, height },
-  }: DimensionChangeEvent) =>
+  }: DimensionChangeEvent) => {
+    // Prevents issue with odd orientation switch when app put in background
+    if (/inactive|background/.test(appState.current)) return;
```
Geoff-Ford, line 33: This is the fix which prevents dimensions being updated when the app goes to the background.

aronshamash (approved):

packages/responsive/src/responsive.tsx
kamiranoff: Ah nice. I wonder if it fixes also a couple of other bugs as well where we can see TCN page resize when it shouldn't.

kamiranoff (approved):

PR #381, opened 2021-03-11
feat: (TNLT-6957) Small Tablet Differentiation

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-6957


packages/article-list/src/article-list-item.js
```diff
@@ -75,13 +75,13 @@ class ArticleListItem extends Component {
 
     return (
       <ResponsiveContext.Consumer>
-        {({ isTablet }) => (
+        {({ isArticleTablet }) => (
```
aronshamash, line 78: We are using `isArticleTablet` on a list page. could it be better to have have `isTablet` for article and list pages and and a specific `isSectionTablet` or `isEditionTablet` for section pages ? I _know_ that is alot of updates...

packages/article-skeleton/__tests__/body-utils/setup-ad.js
```diff
@@ -20,12 +20,15 @@ export default () => {
...
     isTablet: true,
+    isArticleTablet: true,
     narrowContent: false,
   };
 
   describe("setupAd", () => {
     it("should return content untouched if not tablet", () => {
-      expect(setupAd({ ...skeletonProps, isTablet: false })).toEqual(content);
+      expect(
+        setupAd({ ...skeletonProps, isTablet: false, isArticleTablet: false }),
```
aronshamash, line 30: we need both flags ?

packages/article-skeleton/src/styles/inline-newsletter-puff.js
```diff
@@ -117,6 +117,7 @@ const hugeBreakpointStyle = { ...wideBreakpointStyles };
 
 const stylesResolver = {
   [editionBreakpoints.small]: smallBreakpointStyles,
+  [editionBreakpoints.smallTablet]: smallBreakpointStyles,
```
aronshamash, line 120: `smallTablet` slightly breaks the naming convention no? What if some phones end up using this breakpoint? 

`mediumSmall`
`smallPlus`
`compact` 
any better you think ?

packages/responsive/src/calculateResponsiveContext.ts
```diff
@@ -25,7 +27,8 @@ export const calculateResponsiveContext = (
   editionBreakpoint: getEditionBreakpoint(width),
   narrowArticleBreakpoint: getNarrowArticleBreakpoint(width),
   fontScale,
-  isTablet:
+  isTablet: isTablet ? isTablet() : false,
```
aronshamash, line 30: ok so disregard my earlier comment! isTablet makes sense.

packages/responsive/src/calculateResponsiveContext.ts
```diff
@@ -25,7 +27,8 @@ export const calculateResponsiveContext = (
...
+  isArticleTablet:
```
aronshamash, line 31: this one is the more problematic as it is not only used on the article page. I can't think of a better name at the moment.

aronshamash (commented):
good work. huge PR!

packages/article-list/src/article-list-item.js
aronshamash: I have just seen `isTablet` is based on device info so it makes sense to stay..

packages/edition-slices/src/slices/shared/styles.js
```diff
@@ -6,6 +6,9 @@ import {
 } from "@times-components-native/styleguide";
 
 const tabletPaddingHorizontalResolver = {
+  [editionBreakpoints.small]: 0,
+  [editionBreakpoints.smallTablet]: 0,
+  // [editionBreakpoints.smallTablet]: spacing(2),
```
kamiranoff, line 11: left out comment.

packages/article-list/src/article-list-item.js
kamiranoff: I am not sure I understand what `isArticleTablet` means. Is there differences between articles for mobile and articles for tablet in the api?

kamiranoff:
It is a bit hard to review with that many changes, to be honest. There is a lot of repetition in tiles, should that be refactored at some point?
One concern I have is android has a lot of different sizes through multiple devices. It will be hard to cover every edge cases.
 
Anyway, well done, that is a lot of work!

packages/article-list/src/article-list-item.js
Geoff-Ford: @kamiranoff Yes there is difference for article templates. Because we will have "tablet" sections but "mobile" articles on the small 7" Android devices, I had to split this logic up a bit. `isTablet` is now only used in sections and is determined by using native device info (which is based on Portrait width only). This is normally used to decide which slice to render. However `isArticleTablet` is based on a width, regardless of orientation and is now only used in articles.

packages/article-skeleton/__tests__/body-utils/setup-ad.js
Geoff-Ford: Nope, will remove.

packages/article-list/src/article-list-item.js
aronshamash: but the reason I think naming could be better is that `isArticleTablet` is used on list pages.

aronshamash (approved):

packages/edition-slices/src/tiles/tile-ah/styles/index.js
```diff
@@ -8,12 +8,14 @@ import {
 } from "@times-components-native/styleguide";
 
 const headlineFontSizeResolver = {
+  [editionBreakpoints.smallTablet]: 30,
```
kamiranoff, line 11: it seems for many cases `smallTablet` and `medium` are set to the same dimensions. Not for this PR. but I wonder if there is a better way to handle specific instances where `smallTablet` and `medium` are different

kamiranoff (approved):

packages/edition-slices/src/tiles/tile-ah/styles/index.js
Geoff-Ford: Agreed. We have inherited quite a complex handling of different breakpoints for both slice/template selections and also at a style level. Would need quite a large refactor.

PR #382, opened 2021-03-15
fix: (TNLT-7048) SupplementLeadOneAndFourV2 Lead Tile image in iOS landscape

Geoff-Ford:
https://nidigitalsolutions.jira.com/browse/TNLT-7048

* Fixes the landscape layout of the LeadTile image in the SupplementLeadOneAndFourV2 (only used by iOS) when lead asset is a video.

**Before:**
Screenshot 2021-03-15 at 17 29 15

**After:**
Screenshot 2021-03-15 at 17 30 18

Screenshot 2021-03-15 at 17 30 37

### DIFF
--- a/packages/edition-slices/src/tiles/shared/tile-image.tsx
+++ b/packages/edition-slices/src/tiles/shared/tile-image.tsx
@@ -41,6 +41,7 @@ const TileImage: FC<Props> = ({
       <Image
         onLayout={handleOnLayout}
         {...props}
+        style={style}
         key={isPortrait ? "tile-image-portrait" : "tile-image-landscape"}
       />
       {!hideVideoIcon && (


aronshamash (approved):

packages/edition-slices/src/tiles/shared/tile-image.tsx
```diff
@@ -41,6 +41,7 @@ const TileImage: FC<Props> = ({
       <Image
         onLayout={handleOnLayout}
         {...props}
+        style={style}
```
kamiranoff, line 44: shouldnt be snapshots modified with this too?

kamiranoff (approved):

packages/edition-slices/src/tiles/shared/tile-image.tsx
Geoff-Ford: Apparently not. I did run all the `edition-slices` tests and nothing changes. Same on CI below.

PR #390, opened 2021-03-26
feat:(TNLT-6725) Offline screen

marionoratoRN:
This adds the offline screen for the search functionality.

iOS:

https://user-images.githubusercontent.com/41950841/112622719-53735900-8e23-11eb-9b96-0c42e224d010.mov

Android:

https://user-images.githubusercontent.com/41950841/112622748-5ec68480-8e23-11eb-816d-c7e6031a254b.mp4


packages/search/src/offlineMessage.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+import { Text, View } from "react-native";
+import { styles } from "./styles/offlineMessageStyles";
+
+export const OfflineMessage: React.FC = () => {
+  return (
```
kamiranoff, line 6: could remove the return statement here

packages/search/src/search.tsx
```diff
@@ -28,16 +29,29 @@ const getSearchClient = (algoliaConfig: SearchProps["algoliaConfig"]) => {
 };
 
 const Search: FC<SearchProps> = ({ onArticlePress, algoliaConfig }) => {
+  const [isConnected, setIsConnected] = useState<boolean | null>(true);
+  useEffect(() => {
```
kamiranoff, line 33: you could extra this in a reusable hook.

packages/search/src/searchBar/styles/magnifierStyles.ts
```diff
@@ -0,0 +1,9 @@
+import { StyleSheet } from "react-native";
+
+export const styles = StyleSheet.create({
```
kamiranoff, line 3: hum not sure about extracting the styles to new files. I personally like to keep the style in the same file as the component. But its a personal preference

packages/search/__tests__/android/chevron.test.tsx
```diff
@@ -1,5 +1,5 @@
 import TestRenderer from "react-test-renderer";
-import Chevron from "@times-components-native/search/src/search-bar/chevron";
+import Chevron from "@times-components-native/search/src/searchBar/chevron";
```
kamiranoff, line 2: not sure this follows the convention of the other packages. I have personally no preference.

kamiranoff (approved):
Nice one!

packages/search/__tests__/android/chevron.test.tsx
marionoratoRN: I am not sure as well Kevin, that's why I will wait for someone else to approve the PR before merging it. It's just that the kebab-case looks very webby to me. If this is consistent with the project naming convention I am happy to revert

packages/search/src/searchBar/styles/magnifierStyles.ts
marionoratoRN: Same thing for the kebab notation, if this is not consistent with the repo best practice I am happy to revert

packages/search/__tests__/android/chevron.test.tsx
aronshamash: the project started as a React Native Web project and was then split into just a RN repo for use only by the iOS and Android apps. I would go with consistency in this case. I know we are trying to improve the codebase and convert over to typescript when we can which creates inconsistency until it is 100% TS.

packages/icons/src/icons/glass-magnifier.tsx
```diff
@@ -2,7 +2,11 @@ import React, { FC } from "react";
 import Svg, { G, Mask, Path, Rect } from "react-native-svg";
 import { EMaskUnits } from "@times-components-native/icons/src/icons/types";
 
-const GlassMagnifier: FC = () => (
+interface GlassMagnifierProps {
+  color?: string;
+}
+
+const GlassMagnifier: FC<GlassMagnifierProps> = ({ color = "#FFFFFF" }) => (
```
aronshamash, line 9: even this default should come from the styleguide colour palette

packages/search/__tests__/android/chevron.test.tsx
aronshamash: On the other hand if we want to do a conversion of file names to lower camel case as part of a plan to improve the codebase then that is another matter. @kamiranoff @manosim @Geoff-Ford thoughts ?

packages/search/src/searchBar/styles/searchBarStyles.ts
```diff
@@ -0,0 +1,30 @@
...
+    marginVertical: 8,
+  },
+  input: {
+    ...fontFactory({
+      font: "supporting",
+      fontSize: "body",
+    }),
+    lineHeight: 22,
+    padding: 4,
+    paddingLeft: 30,
```
aronshamash, line 24: multiple of `spacing`

packages/search/src/searchList/styles/searchListItemByLineStyles.ts
```diff
@@ -0,0 +1,10 @@
+import styleguide from "@times-components-native/styleguide/src/styleguide";
+import { StyleSheet } from "react-native";
+
+const { spacing } = styleguide();
+
+export const styles = StyleSheet.create({
+  byLineContainer: {
```
aronshamash, line 7: `bylineContainer` typo

packages/search/__tests__/android/chevron.test.tsx
kamiranoff: hum. I do like the Java paradigm where the file name is identical that the exported class so when you search for a class you can find it easily, but that might not really apply here.
I think what is more important to have named exports to make sure we can navigate in the IDE without searching for too long.

PR #406, opened 2021-04-21
fix: (TNLT-7334) android crashes

aronshamash:
- Fix android crashing after multiple rotations.

### DIFF
--- a/packages/tooltip/styles/index.js
+++ b/packages/tooltip/styles/index.js
@@ -126,13 +126,11 @@ const generateStyles = (options) => {
       flexDirection: "column-reverse",
       justifyContent: "space-between",
     };
-    containerPlacementStyles = {
-      ...{
-        position: "relative",
-        marginTop: spacing(4),
-      },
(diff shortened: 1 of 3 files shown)


packages/tooltip/styles/index.js
```diff
@@ -126,13 +126,11 @@ const generateStyles = (options) => {
       flexDirection: "column-reverse",
       justifyContent: "space-between",
     };
-    containerPlacementStyles = {
-      ...{
-        position: "relative",
-        marginTop: spacing(4),
-      },
-      ...containerPlacementStyles,
-    };
```
jonnyandrew, line 135: Nice find @aronshamash! Is this the recursion from the stack trace?

aronshamash: I _think_ so

kamiranoff: Nice find @aronshamash!

kamiranoff (approved):

PR #410, opened 2021-04-22
fix: TNLT-7403 TNLT-7420 Keyboard on empty search screen + wrong fonts

marionoratoRN:
This PR fixes the keyboard issue on the empty search screen and some fonts that were not consistent with the design

### DIFF
--- a/packages/search/src/search-list/styles/search-list-empty-state-styles.ts
+++ b/packages/search/src/search-list/styles/search-list-empty-state-styles.ts
@@ -4,6 +4,7 @@ import {
   fontFactory,
   spacing,
 } from "@times-components-native/styleguide";
+import { GillSansMTStdMedium } from "../../../../../app/utils/fonts";
 
 export const styles = StyleSheet.create({
   listEmptyTitle: {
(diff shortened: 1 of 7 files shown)


packages/search/src/search-list/styles/search-list-empty-state-styles.ts
```diff
@@ -4,6 +4,7 @@ import {
   fontFactory,
   spacing,
 } from "@times-components-native/styleguide";
+import { GillSansMTStdMedium } from "../../../../../app/utils/fonts";
```
aronshamash, line 7: Can you add this font to https://github.com/newsuk/times-components-native/blob/674a0174a6a89bc7fdca4e4d3922342d385a6ad7/packages/styleguide/src/fonts/fonts.ts and import into this style file in the same way as other fonts ?

packages/search/src/search-list/styles/search-list-item-snippet-styles.ts
```diff
@@ -1,8 +1,10 @@
 import { StyleSheet } from "react-native";
 import { colours } from "@times-components-native/styleguide";
+import { TimesModern } from "../../../../../app/utils/fonts";
```
aronshamash, line 3: can you import in same way as other fonts ?

aronshamash (changes requested):

aronshamash (approved):

PR #412, opened 2021-04-26
fix: (TNLT-7472) Pass isArticleTablet in all article templates

aronshamash:
- Fixes https://nidigitalsolutions.jira.com/browse/TNLT-7472
- `isArticleTablet` was not being passed to `ArticleSkelton` in a couple of templates.


reactnativecoder (approved):
LGTM

PR #416, opened 2021-04-27
fix: (TNLT-6802) full screen video ads

aronshamash:
- fix issue in ios where videos would play full screen by default.


reactnativecoder (approved):
LGTM

PR #419, opened 2021-04-30
chore: Update codeowners

aronshamash:
- Remove @kamiranoff and myself


kamiranoff (approved):

marionoratoRN (approved):
I feel a bit alone, so I cannot write LGTM. I will approve it, but it does not feel really good

PR #424, opened 2021-05-14
fix: Article end tracking event

reactnativecoder:
This PR changes the analytics event that gets triggered on article bottom reached.

As much as I understood wrapping the ArticleEndTracking with withTrackingContext, triggers an event rather than view event. So this fix should work.

Due to the fact that in the rest of the project these kind of events are handled in native app (TheTimesProjectD -> AnalyticsService), I can't be 100% sure that withTrackingContext is right way to handle it.


mrdanb (commented):
Hi @marionoratoRN – 
I just tried running this branch locally in the iOS app and I can no longer see `"view end"` events coming through when scrolling to the end of an article. This was working ok before (at least on iOS).

I'm a bit confused what the issue is tbh. I'm assuming it is for this ticket https://nidigitalsolutions.jira.com/browse/TNLT-6872 ?

If possible could you clarify what we are trying to fix? Thanks!

newsuk-lee-daffen:
@petermason @smccamley This is pretty old. Can we close it?

PR #426, opened 2021-05-27
feat: Add analtyics to Search

petermason:
Add custom tracking events for the Search feature
ref: https://nidigitalsolutions.jira.com/browse/TNLT-7407
* Search page
* Search results page
* Offline page
* No results page

Note: This PR uses the native bridge Track function directly for simplicity. This is used in preference to withTrackingContext as this code is deprecated and uses unecessary abstraction.

example

```
import { NaviteModules } from 'react-native';

const { track } = NativeModules.ReactAnalytics;

track(trackingData)
```

### DIFF
--- a/packages/search/src/search-list/search-list-empty-state.tsx
+++ b/packages/search/src/search-list/search-list-empty-state.tsx
@@ -1,30 +1,42 @@
-import React from "react";
+import React, { useEffect } from "react";
+import { NativeModules } from "react-native";
 import { Text, Image, KeyboardAvoidingView, Platform } from "react-native";
 import { styles } from "./styles/search-list-empty-state-styles";
 import { ImageIcons } from "@times-components-native/icons/src/icons/imageIcons";
+import { TTrackingData } from "../types";
+
+const { track } = NativeModules.ReactAnalytics;
 
 interface SearchListEmptyStateProps {
   message: string;
   title: string;
   icon?: string;
+  trackingData: TTrackingData;
 }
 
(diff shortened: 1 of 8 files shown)


packages/search/src/search-list/search-list-empty-state.tsx
```diff
@@ -1,30 +1,42 @@
-import React from "react";
+import React, { useEffect } from "react";
+import { NativeModules } from "react-native";
 import { Text, Image, KeyboardAvoidingView, Platform } from "react-native";
 import { styles } from "./styles/search-list-empty-state-styles";
 import { ImageIcons } from "@times-components-native/icons/src/icons/imageIcons";
+import { TTrackingData } from "../types";
```
newsuk-lee-daffen, line 6: Did you intend to have a double T in `TrackingData`?

petermason: Sorry, yes habit, naming convention T === type I === Interface.

newsuk-lee-daffen: Ah, okay, makes sense. However, none of the other types we have in `types.ts` use the `T` prefix. Can I be a pain and suggest we remove it for consistency?

petermason: Updated, I also moved the type definition to the the top level types director to make it generically available to all modules that may use tracking

newsuk-lee-daffen (approved):

jonnyandrew:
@petermason is this still planned or can we close?

petermason:
not required

PR #431, opened 2021-06-10
fix: remove add from article render for test

petermason:
a test, removes adds from article render


jonnyandrew:
@petermason I'm assuming we can close this one?

petermason:
not required

PR #439, opened 2021-06-29
fix: article end tracking

petermason:


smccamley (approved):
actionName makes sense ;-) link

PR #448, opened 2021-07-20
fix: tnlt 8061 broken text

petermason:
Hot fix P1: Removes an wrapping View component that was causing Text to break

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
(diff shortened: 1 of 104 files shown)


jonnyandrew (approved):
Out of interest why would we have needed the wrapping view originally?

petermason:
> Out of interest why would we have needed the wrapping view originally?

this was orignially used for debugging the component as should have been removed!

PR #449, opened 2021-07-23
fix: tnlt 7938 fix times domain ads

petermason:
adjust ad logic to allow ads to resolve inside times.co.uk domain. previously this was not allowed and the side effect of this was for the ad to actually load inside the ad contianer. see --> https://nidigitalsolutions.jira.com/browse/TNLT-7938


jonnyandrew (commented):
Might this have an effect on other content that should load inside the webview? I'm thinking some puffs/interactives might need this behaviour

petermason:
> Might this have an effect on other Times content that should load inside a webview? I'm thinking some puffs/interactives might need the existing behaviour

The fix shouldn't affect the content (the ad) that's loaded into the webview but handle the click event. It checks that the ad is pointing to the host domain.

jonnyandrew (approved):

PR #450, opened 2021-07-23
fix: tnlt 6777 is tablet checks

petermason:
- revert dimension change listener to previous working solution
- remove return code that exited android devices before dimension update. 
The combination of the above two steps appears to fix this issue. (subject to aditional testing)


packages/utils/src/dimensions-util.js
```diff
@@ -7,10 +18,45 @@ export const getDimensions = (
...
 
 export const removeDimensionsListener = (type, handler) => {
-  Dimensions.removeEventListener(type, handler);
+  if (Platform.OS === "android") {
+    eventEmitter.removeListener("configurationChanged", handler);
+  } else {
+    Dimensions.removeEventListener(type, handler);
+  }
 };
+// import { Dimensions } from "react-native";
```
jonnyandrew, line 47: I would remove the commented out code since it is already available in the Git history. Considering how tricky this issue is, it could be useful to add some documentation here about the workaround and the issues with Dimensions library.

jonnyandrew (approved):

PR #475, opened 2021-09-15
fix(TNLT-8224): Incompatible `react-native-webview` build

jonnyandrew:
#### Description
The native Android TCN project (under`lib/android`) is a Gradle multi-module project. The project consists of one core module `lib/android/xnative` and some linked subprojects that are linked from within `node_modules` in the root project (see `lib/android/settings.gradle`). These subprojects are built and deployed alongside the core TCN artifact (`times-xnative`) and are hence interdependent on the whole `xnative` project build.

However, currently any TCN release will always deploy these dependencies and if the underlying Node dependency versions did not change, **the existing artifacts are overwritten in the remote repository**. This can break compatibility between the dependencies when new TCN releases are made.

To fix this, I have updated the naming and versioning of those React Native dependencies to reflect this coupling. Now, each new deployment of TCN library will also generate a new artifact with a new version for each of the linked modules.

This will not fix the dependencies that have been overwritten in the past, but it will prevent this type of problem happening in the future.

**Artifacts deployed before**
- `uk.co.thetimes:times-xnative:<VERSION_X>`
- `react-repo:react-native-webview:<VERSION_Y>`
- etc.

**Artifacts deployed now**
- `uk.co.thetimes:times-xnative:<VERSION_X>`
- `uk.co.thetimes:xnative-webview:<VERSION_X>` <-- **note: now same version as `times-xnative`**
- etc.

##### jFrog deployment
Screenshot 2021-09-15 at 14 08 23

#### Checklist
 - [ ] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet
 
I have tested locally that these new dependencies are resolved properly using TCN `6.0.2-beta0`.


abhisheknewscorp (approved):

abhisheknewscorp (approved):
**Praise:** It is a nice catch. Thanks for addressing it

PR #472, opened 2021-09-09
fix(TNLT-8196): invisible webview

smccamley:
#### Checklist
 - [ ] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Android emulator
 - [ ] Tested on Tablet

PR #478, opened 2021-09-17
fix: NO JIRA - revert webview versions

petermason:
NO JIRA

#### Description

Revert webview components to earlier version to prevent blank interactive content


jonnyandrew:
Closing in favour of https://github.com/newsuk/times-components-native/pull/472

PR #494, opened 2021-10-25
fix: TNLT-8277 Update fronts layout to be responsive

petermason:
#### Description

WIP

This fix focuses primarily on Fronts but introduces logic for a new responsive design pattern. The logic for this is based on the device screen width combined with the fontScale value, as defined by the user in the device accessibility settings.

Note: Media Queries are not supported by RN Styles, so here I've created a media query hook. There are some npm modules available but they don't take font scale into consideration.

New MediaQuery Hook

//packages/hooks/media-query

```
import { useWindowDimensions } from "react-native";

export type TMediaQuerySize =
  | "EXTRA SMALL"
  | "SMALL"
  | "MEDIUM"
  | "LARGE"
  | "EXTRA LARGE"
  | "EXTRA EXTRA LARGE";

function MediaQuery(): TMediaQuerySize {
  const { width, fontScale } = useWindowDimensions();
  const breakpoint = width / fontScale;

  let size: TMediaQuerySize = "EXTRA SMALL";

  // 480px, 768px, 1024px, and 1280px

  if (breakpoint >= 480 && breakpoint < 600) {
    size = "SMALL";
  }
  if (breakpoint >= 600 && breakpoint < 768) {
    size = "MEDIUM";
  }
  if (breakpoint >= 768 && breakpoint < 1024) {
    size = "LARGE";
  }
  if (breakpoint >= 1024 && breakpoint < 1280) {
    size = "EXTRA LARGE";
  }
  if (breakpoint >= 1280) {
    size = "EXTRA EXTRA LARGE";
  }

  return size;
}
```
Note: the above is a first draft as a suggestion to UX&D and uses common breakpoints

To work with the media query I've introduced two new generic layout components that respond to the various breakpoints

example responsive container: c-100-50.tsx

```
import React, { ReactNode } from "react";
import { View } from "react-native";

import { MediaQuery } from "@times-components-native/hooks";

// A generic responsive Container
// EXTRA SMALL = 100
// > EXTRA SMALL = 50

interface IProps {
  children: ReactNode;
}

function CHundredFifty({ children }: IProps) {
  const screenSize = MediaQuery();

  const getWidth = () => {
    switch (screenSize) {
      case "EXTRA SMALL":
        return "100%";
      default:
        return "50%";
    }
  };

  const width = getWidth();

  return (
    <View
      style={{
        width,
      }}
    >
      {children}
    </View>
  );
}

export default CHundredFifty;
```
The above container simply resizes based on the current size provided by the Media Query Hook. See also //packages/layouts/containers/c-100-50-25.tsx

The naming convention of these components is show the possible widths. i.e c-100-50.tsx can be 100% wide and 50%.

The main work included in this PR is to apply these new layout tools to the Front section, so this means modifying some existing templates and applying the new responsive logic.

#### default
image

#### small
image

#### extra small
image

#### default
image

same priniciples apply in landscape

e.g
image

### Old Front Templates

Fronts has a number of templates. So far I have modified
* FrontLeadTwo
* FrontLeadOneAndOne

Originally these layout used a fixed layout and didn't scroll. This broke on devices like the new iPad Mini. I've introduced a flexible scroll that allows scrolling when the content overflows. This happens when the font size is scaled.

example: ipad mini with large font scaling applied creates a scrolling single column layout. Note fonts not actually scaled in screen shot (font scaling is currently disabled)

image

#### Checklist
 - [ ] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet


packages/in-todays-edition/styles/index.js
```diff
@@ -1,254 +1,63 @@
...
-  },
-  itemCTAText: {
-    ...sharedLandscapeStyles.itemCTAText,
-    fontSize: 14,
-  },
-};
-
-const styles = {
-  portrait: {
-    744: {
```
jonnyandrew, line 143: 🤩

packages/layouts/containers/c-100-50-25.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { ReactNode } from "react";
+import { View } from "react-native";
+
+import { MediaQuery } from "@times-components-native/hooks";
+
+// A generic responsive Container
+// EXTRA SMALL = 100
+// SMALL = 50
+// > SMALL = 25
```
jonnyandrew, line 9: Am I right that this a single 'item' in a row or column? What happens if we compose more than one together in a full width container? Will the wrapping just work?

packages/layouts/containers/c-100-50-25.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+type StyleObject = { [key: string]: string | number };
+
+interface IProps {
+  children: ReactNode;
+  responsive?: boolean;
+  style?: StyleObject;
+}
+
+function CHundredFiftyTwentyFive({
```
jonnyandrew, line 19: At first I was a bit confused by the naming convention (but I am not too sure either how to improve it!). Maybe something that references the breakpoints and/or uses `_` separators (e.g.`Col_25_S50_Xs100`) could be clearer?

jonnyandrew (commented):
Looks really nice so far!

packages/typography/headers/h1.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { MediaQuery } from "@times-components-native/hooks";
+
+interface IProps {
+  children: ReactNode;
+}
+
+function H1({ children }: IProps) {
+  const screenSize = MediaQuery();
+
+  const getFontSize = () => {
```
smccamley, line 12: It looks like this H1 component is going to be rendered. 
If so we shouldn't have functions that could not depend on props inside the rendered components

I would pull this function outside of the H1 function and make screenSize a prop

packages/hooks/media-query/index.tsx
```diff
@@ -0,0 +1,38 @@
+import { useWindowDimensions } from "react-native";
+
+export type TMediaQuerySize =
+  | "EXTRA SMALL"
+  | "SMALL"
+  | "MEDIUM"
+  | "LARGE"
+  | "EXTRA LARGE"
+  | "EXTRA EXTRA LARGE";
```
jonnyandrew, line 9: Did UX have any thoughts on the breakpoints we'd like to support? I like the simplicity of the [three size classes in the new Google standard](https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes#window_size_classes).

image

jonnyandrew (commented):
Looks great! Just a couple of questions.

packages/edition-slices/src/slices/frontleadoneandone/index.js
```diff
@@ -1,64 +1,119 @@
...
+                onLayout={(event) => {
+                  setLeftColWidth(event.nativeEvent.layout.width);
+                }}
+              >
+                <TileAFront
+                  onPress={onPress}
+                  tile={lead}
+                  tileName="lead"
+                  orientation={"portrait"}
+                  colWidth={leftColWidth}
```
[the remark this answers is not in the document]
petermason: This is a good point, I'm all for removing complexity (a lot has already gone). Let me take a look and see if I can think of a better way to do this

packages/hooks/media-query/index.tsx
petermason: I like this too. My current breakpoints are there until UX define them in the next phase, which should be soon

smccamley (approved):

petermason:
stale work

PR #499, opened 2021-11-03
fix(TNLT-8060): ad-height-dynamic

smccamley:
#### Description
This ticket was about increasing the height of Dianomi ads so the 2nd ad was not cut off. The suggestion was to set a hard coded height as it was believed that the ads were a fixed height. It turned out that
* Ads were not fixed height
* Ad heights were scale relative to the device width
* The number of ads that are shown are not always two and we do not always have control over this number
* The ads are displayed inside an iFrame inside a Webview

The only way to know how many ads are loaded into the Webview was to understand the structure of the web content and pass this information up to react native.
- I have used cheerio to interpret the HTML
- and tried to simplify the process of understanding the height calculations
- There is also a catch all to prevent this logic from effecting tablet where there is always one row of ads.

#### Checklist
 - [x] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [x] Tested on iOS simulator
 - [x] Tested on Tablet


packages/ad/src/styles/index.js
```diff
@@ -2,9 +2,13 @@ import styleguide from "@times-components-native/styleguide";
 import { StyleSheet, Dimensions } from "react-native";
 
 const { colours, fontFactory, spacing, fonts } = styleguide();
-
 const { width } = Dimensions.get("screen");
-const isNarrowScreenDevice = width < 600;
+
+const AD_WIDTH_HEIGHT_RATIO = 414 / 340;
```
jonnyandrew, line 7: How did you arrive at this ratio? Won't the ratio change depending on the ad height and number of ads?

packages/ad/src/styles/index.js
```diff
@@ -2,9 +2,13 @@ import styleguide from "@times-components-native/styleguide";
...
+export const TABLET_AD_HEIGHT = 350;
+
+export const getAdHeightForNAds = (numberOfAds, adWidth = width) =>
+  numberOfAds * (adWidth / AD_WIDTH_HEIGHT_RATIO);
```
jonnyandrew, line 11: Do these fixed values handle different system font size settings?

packages/ad/src/SponsoredAd.tsx
```diff
@@ -0,0 +1,149 @@
...
+const SponsoredAd: React.FC<SponsoredAdProps> = ({ numberOfAds = 4 }) => {
+  const [height, setHeight] = useState<number>(0);
+  const webview = useRef<WebView>(null);
+
+  const handleOnMessage = (event: WebViewMessageEvent) => {
+    if (event.nativeEvent.data && !isTablet()) {
+      const eventData = JSON.parse(event.nativeEvent.data);
+      if (eventData && eventData.adHtml) {
+        const $ = cheerio.load(eventData.adHtml);
+        const numberOfAdsInHtml = $("div[id^='dianomi_ad_']").length;
```
jonnyandrew, line 115: This looks like an internal Dianomi `id` meaning it could change any time and break the layout.

jonnyandrew (commented):
I'm curious why we couldn't use `AutoHeightWebView` for this like we do with interactives?

packages/ad/src/styles/index.js
smccamley: Unfortunately The Dianomi ad font size does not change with the system font size
I didn't investigate if we could pass the font size down from react -> webview -> dynamically loaded iframe, 
I guessed with the font size work going on that this might be another task.

packages/ad/src/SponsoredAd.tsx
smccamley: It is part of the ID and it could break if Dianomi change their html. I tried lots of ideas to solve this one, and I agree it's a little hacky. The issue is that as the content is in nested unrendered iframe inside a webview, I couldn't rely on the frames to reliably tell me what their height, so I had to rely on the html. So I made a judgement that it's likely the least likely part of the html to change. And in the event where it fails, it fails gracefully showing no big empty whitespace section and setting the height to 0.

packages/ad/src/styles/index.js
smccamley: This is just a ratio of the width to the height of a single ad. The height of the ad container will change based on the number of ads (that's calculated dynamically using the getAdHeightForNAds function)

jonnyandrew: I see - so is this a standard ad size that we are requesting (i.e. the ratio won't change in future)?

packages/ad/src/styles/index.js
jonnyandrew: Ok then I agree that is outside the scope of this

packages/ad/src/SponsoredAd.tsx
jonnyandrew: Did you check whether `AutoHeightWebView` works for this use case?

If it doesn't, good that it fails gracefully 👍

packages/ad/src/SponsoredAd.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { WebViewNavigation } from "react-native-webview/lib/WebViewTypes";
+import cheerio from "cheerio";
+
+import styles, {
+  getAdContainerHeightForNAds,
+  TABLET_AD_HEIGHT,
+} from "./styles";
+
+interface WebviewHeightCallbackSetupProps {
+  window: any;
```
petermason, line 14: is it not possible to type these?

packages/ad/src/SponsoredAd.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  Linking.openURL(e.url);
+  return false;
+};
+
+const scriptToInject = `window.postMessage = function(data) {
+  window.ReactNativeWebView.postMessage(data);
+};(${webviewHeightCallbackSetup.toString()})({window,MutationObserver});`;
+
+const numberToContextID: Record<number, string> = {
```
petermason, line 98: where do these values come from? and are they fixed?

packages/ad/src/SponsoredAd.tsx
```diff
@@ -0,0 +1,152 @@
...
+  2: "250",
+  3: "244",
+  4: "243",
+};
+
+interface SponsoredAdProps {
+  numberOfAds?: number;
+}
+
+const SponsoredAd: React.FC<SponsoredAdProps> = ({ numberOfAds = 4 }) => {
```
petermason, line 109: default to 4 ads? Is this requirement defined anywhere?

packages/ad/src/SponsoredAd.tsx
```diff
@@ -0,0 +1,152 @@
...
+        const $ = cheerio.load(eventData.adHtml);
+        const numberOfAdsInHtml = $("div[id^='dianomi_ad_']").length;
+        setHeight(getAdContainerHeightForNAds(numberOfAdsInHtml));
+      }
+    } else {
+      setHeight(TABLET_AD_HEIGHT);
+    }
+  };
+
+  const contextId = numberToContextID[numberOfAds] || numberToContextID[4];
```
petermason, line 126: logic? number of ads has a default value of 4

packages/ad/src/SponsoredAd.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+const webviewHeightCallbackSetup = ({
+  window,
+  MutationObserver,
+}: WebviewHeightCallbackSetupProps) => {
+  "show source";
+  const document = window.document;
+  window.eventCallback = (type: any, content: any) => {
+    window.ReactNativeWebView.postMessage(
+      JSON.stringify({ rar: "the thing", type, content }),
```
petermason, line 26: what's 'the thing'

packages/ad/src/SponsoredAd.tsx
```diff
@@ -0,0 +1,152 @@
...
+  MutationObserver: any;
+}
+
+const webviewHeightCallbackSetup = ({
+  window,
+  MutationObserver,
+}: WebviewHeightCallbackSetupProps) => {
+  "show source";
+  const document = window.document;
+  window.eventCallback = (type: any, content: any) => {
```
petermason, line 24: use of any

packages/ad/src/SponsoredAd.tsx
```diff
@@ -0,0 +1,152 @@
...
+    window.ReactNativeWebView.postMessage(
+      JSON.stringify({ rar: "the thing", type, content }),
+    );
+  };
+  window.addEventListener("error", (ev: any) => {
```
petermason, line 29: use of any here, but know properties used below

packages/ad/src/SponsoredAd.tsx
smccamley: This is old logic, could clean up, whomever originally wrote this component assumed default was tablet. I could change this

packages/ad/src/SponsoredAd.tsx
smccamley: This is also old logic, they are the ad placement types from Dianomi I believe. I could rename.

packages/ad/src/SponsoredAd.tsx
smccamley: No, you can't import types from the dom. Typescript makes certain basic assumptions about where a project runs and so it would be a risky change to the way TS works for our project. Jonathan and I talked about maybe moving some of this JS (browser style) code elsewhere and into it's own repo or something similar, but that would also add complexity as it's only used here.

packages/ad/src/SponsoredAd.tsx
smccamley: Unfortunately I couldn't make it work just with AutoHeightWebview, it looks to be a bug / a tricky use case that's similar to the video issue you see in their issues page

packages/ad/src/SponsoredAd.tsx
smccamley: The content can be any, i've changed the type on the `type` to string

packages/ad/src/SponsoredAd.tsx
smccamley: it's gone :-)

packages/ad/src/SponsoredAd.tsx
smccamley: This relates to the above comment, DOM code in a non dom project and not able to import proper types for this.

packages/ad/src/SponsoredAd.tsx
smccamley: Change the old logic and cleaned up

packages/ad/src/SponsoredAd.tsx
smccamley: I've changed it

packages/ad/src/SponsoredAd.tsx
smccamley: I've renamed this so it's easier to understand.

packages/ad/src/SponsoredAd.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  Linking.openURL(e.url);
+  return false;
+};
+
+const scriptToInject = `window.postMessage = function(data) {
+  window.ReactNativeWebView.postMessage(data);
+};(${webviewHeightCallbackSetup.toString()})({window,MutationObserver});`;
+
+const placementReferenceByAdCount: Record<number, string> = {
```
petermason, line 96: Where do these values come from and what do they represent?

smccamley: They are Dianomi ad placement numbers and are not part of code i've changed, simply copied and renamed slightly to improve readability

packages/ad/src/styles/index.js
smccamley: Yes, these sizes are set in Dianomi based on the ad placements and don't change unless we ask for them to change

packages/ad/src/SponsoredAd.tsx
petermason: Maybe make them constansts and write some comments so it clear what they are and where they come from. Bit of a mystery otherwise.

smccamley: Makes sense, that's updated

jonnyandrew (approved):

petermason (approved):

PR #501, opened 2021-11-04
fix: TNLT-8268 Article view should be tablet layout on ipad mini

smccamley:
#### Description

In this PR I have decreased the minimum table size breakpoint to 700. The reason for this is that the breakpoint was previously set at the smallest size of an existing iPad. The smallest size of an iPad may have been wise in the past but a breakpoint is encapsulating the idea of a fluid layout and as the apple devices are few and of fixed sizes, it made sense to use the breakpoints set out in the (Android docs)[https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes#window_size_classes]. So it now covers the iPad mini case.

Also it seems that there was the option for the native side to override the breakpoints, from what I can see, this is not being used. So I have removed this code.

Finally there was a potential bug where the code expected a breakpoint to be returned, but it was not. For now I have added a temporary relative value in place here and I think it's best to look at the breakpoint work again once Peter's font scaling work has been merged in.

Important: This feature depends on an update to the iOS app where the hardcoded breakpoints have changed.

#### Checklist
 - [ ] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [x] Tested on iOS simulator
 - [x ] Tested on Android emulator
 - [x] Tested on Tablet


packages/styleguide/src/breakpoints/index.js
```diff
@@ -8,12 +6,7 @@ import widths, {
   narrowArticleWidths,
 } from "./index.shared";
 
-const config = (NativeModules || {}).ReactConfig;
-
 const getEditionBreakpoint = (width) => {
-  if (config && config.breakpoint) {
-    return config.breakpoint;
-  }
```
jonnyandrew, line 16: Thanks for cleaning up this unused code.

packages/image/src/safeAreaView/react-native-safe-area.js
```diff
@@ -16,7 +16,7 @@ const X_WIDTH = 375;
 const X_HEIGHT = 812;
 const XSMAX_WIDTH = 414;
 const XSMAX_HEIGHT = 896;
-const PAD_WIDTH = 768;
+const PAD_WIDTH = 744;
```
petermason, line 19: based on you changes below, should this be 700?

smccamley: yep

packages/styleguide/src/breakpoints/index.shared.ts
```diff
@@ -8,14 +8,18 @@ const editionBreakpoints = {
...
   wide: 1024,
 };
 const editionMaxWidth = editionBreakpointWidths.huge;
 const sliceContentMaxWidth = 1180;
 
 const narrowArticleWidths = {
+  small: {
+    container: 553,
+    content: 545,
+  },
```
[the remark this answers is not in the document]
smccamley: Yes, this is the bug fix, previously some devices were referencing the small breakpoint and nothing existed. I couldn't see any issue with small/mobile layout this was tested in lots of devices.

petermason (approved):

jonnyandrew (approved):

PR #508, opened 2021-11-19
chore(TNLT-8106): Add analytics to track search result clicks

jonnyandrew:
#### Description
Add an analytics event to track clicks of search result items.

##### To do
- [x] Include article data in the tracking call (`article_parent_name`, `other_details`, `article_name`) if possible (see [spec](https://docs.google.com/spreadsheets/d/1G1Uoi89_Y-ivezNhzMt1hZL1jxGDHJMJGYlr4cEYTgc/edit#gid=193578488)).
- [x] Add a test to check the search result list component fires the tracking event when clicked

#### Screenshots
N/A, no visual changes.

#### Checklist
 - [ ] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet


packages/search/src/analytics/search-analytics.ts
```diff
@@ -0,0 +1,93 @@
...
+const trackNative = NativeModules.ReactAnalytics.track;
+
+export const trackSearchResultsPageView: () => void = () => {
+  const attrs: PageTrackingAttributes = {
+    eventTime: new Date(),
+    pageName: "search results",
+    pageSection: "search",
+  };
+
+  const trackingData = createSearchTrackingData({ attrs: attrs });
```
petermason, line 18: should the type attrs be PageTrackingAttributes?

packages/search/src/analytics/search-analytics.ts
```diff
@@ -0,0 +1,93 @@
...
+  attrs: TrackingAttributes;
+};
+
+function createSearchTrackingData({
+  object = "Search",
+  action = "Viewed",
+  component = "Search",
+  attrs,
+}: CreateSearchTrackingDataParameters): TrackingData {
+  return {
```
petermason, line 87: can use shorthand here return {object, action, attrs}. You can do this as the prop and value have the same name

packages/search/src/search-bar/search-bar.tsx
```diff
@@ -24,109 +27,111 @@ type SearchBarComponentProps = PropsWithChildren<SearchBoxProvided> & {
...
-            <View style={styles.inputContainer}>
-              <View style={styles.magnifierTextWrapper}>
-                {isIOS ? (
-                  <Magnifier
-                    style={styles.iconStyle}
-                    color={
-                      !isConnected
-                        ? colours.functional.offlineSearchText
-                        : undefined
+    useEffect(() => {
```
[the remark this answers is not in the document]
smccamley: I didn't `I think it was like that when I got there` but now removed :-)

petermason (approved):

PR #520, opened 2021-12-17
feat: (TNLT-8804) Feature/new flags

smccamley:
#### Description
Just adding two more flags, the only new code really is the new color that has been pulled from the website's implementation of the breaking and live flags.
https://www.staging-thetimes.co.uk/article/48891d64-4bb9-11ec-9bc6-5abf98655bc4

#### Checklist
 - [ ] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet


packages/article-flag/src/article-flag.js
```diff
@@ -90,6 +91,10 @@ const ArticleFlags = ({ flags, longRead, color, style, withContainer }) => {
     EXCLUSIVE: <ExclusiveArticleFlag color={color} />,
     SPONSORED: <SponsoredArticleFlag color={color} />,
     LONGREAD: <LongReadArticleFlag color={color} />,
+    LIVE: <ArticleFlag color={colours.functional.darkRed} title="live" />,
+    BREAKING: (
+      <ArticleFlag color={colours.functional.darkRed} title="breaking" />
+    ),
```
jonnyandrew, line 97: The other flags seem to have a default color but the color is still configurable via the `color` prop. Shouldn't we follow that pattern for the new flags as well?

smccamley: I felt the opposite, we shouldn't be adding in configurable options unless we have to. There is so much `maybe` in this codebase that we have no idea if it's used. I feel it's best we simplify it.

jonnyandrew (approved):

PR #529, opened 2021-12-24
chore(TNLT-8774): Remove old scale references

jonnyandrew:
#### Description
Remove old and unused scale references that may cause confusion.

#### Screenshots 
N/A

#### Checklist
 - [ ] Unit tests
 - [x] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet


DeepfriedDev (approved):
LGTM

packages/article-skeleton/src/article-body/article-body-row.js
```diff
@@ -43,12 +43,10 @@ const ArticleBodyRow = ({
   narrowContent,
   onParagraphTextLayout,
 }) => {
-  const { fontScale } = Dimensions.get("window");
```
petermason, line 46: This value (fontScale) is the value provided by the device and not the value set currently in settings. We actually need this one, although we will need to update the styleFactory to support min and max values.

packages/article-skeleton/src/styles/article-body/shared.js
```diff
@@ -15,8 +10,6 @@ const sharedStyles = ({
     }),
     color: colours.functional.black,
   };
-  defaultFont.fontSize *= fontScale;
```
petermason, line 18: mentioned above. The fontScale value here is the one set by the user in the app settings. However, here we need to support min/max scale values

PR #532, opened 2022-01-11
chore(NO-JIRA): Version Bump 7.1.0

smccamley:
#### Description
Version bump


jonnyandrew (commented):
It looks like the change to be included is a feature (https://github.com/newsuk/times-components-native/pull/531) so I think this next version should be 7.1.0

jonnyandrew (approved):

PR #533, opened 2022-01-11
chore: font scaling - max / min font size

petermason:
# Max / Min Font Scale

## [TNLT-8770](https://nidigitalsolutions.jira.com/browse/TNLT-8770)
and
#### Description
* Restritct max font scale value to 2 (200%)
* Restrict min font scale value to 0.75 (75%)

#### Screenshots 

MAX
image

MIN
image

#### Checklist
 - [ ] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet


jonnyandrew (commented):
There are a lot of places where the min/max scales are configured, making it hard to change. Should we either

1. create two global constants that all `Text`s can reference, or
2. subclass `Text` (e.g. `TimesText`) and use that everywhere?

Also, taking a step back, do we really need to apply this restriction to all these text views? Or could we pick and choose those which are causing issues when scaled to the extreme?

PR #536, opened 2022-01-19
fix(TNLT-8894): prevent gap and ad slicing on error or zero height ad

petermason:
#### Description

Previous code adds additional spacing regardless of whether ad has correctly loaded and provided height.

#### Screenshots

With ad:
image

Missing ad - correct spacing between articles
image

#### Checklist
 - [ ] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet


jonnyandrew (approved):
I'm wondering if it's possible to avoid rendering this component altogether if there is no ad to be shown?

PR #544, opened 2022-03-04
chore: NO-JIRA patch release 7.1.5

petermason:
NO JIRA

#### Description

patch release 7.1.5

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "7.1.4",
+  "version": "7.1.5",
   "description": "A collection of react-native components for The Times and Sunday Times",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "times-components-native",
   "private": true,
-  "version": "7.1.4",
+  "version": "7.1.5",
```
jonnyandrew, line 4:
```suggestion
  "version": "7.2.0",
```

jonnyandrew (commented):
Should this not be 7.2.0 given we've added a new feature?

package.json
petermason: Good point - killing this one

PR #567, opened 2022-04-01
fix(TNLT-8932): pass ReadArticles into section

michalNews:
#### Checklist
 - [ ] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet


petermason (approved):

packages/pages/src/section/section.js
```diff
@@ -35,10 +35,18 @@ const onPuzzlePress = ({ id, title, url }) =>
 class SectionPage extends Component {
   constructor(props) {
     super(props);
+
+    const existingReadArticles =
+      props &&
+      props.readArticles.map((articleId) => ({
```
MSampsonNews, line 41: If read articles is undefined this would throw an undefined error wouldn't it?
"cannot read map of undefined"

Also does proptypes need updating?

michalNews: I added proptypes, since I am passing default value I assume this doesn't need an additional condition 
`props && props. readArticles &&  props.readArticles.map((articleId)`

MSampsonNews: Sweet, I'm happy with that!

MSampsonNews (approved):

petermason (approved):

PR #660, opened 2022-08-12
Feature/tnlt 9126 ad lib and sourcpoint implementation

petermason:
#### Description

Replace current ad config with new Ad Library


packages/ad/__tests__/ios/dom-context.ios.js
```diff
@@ -1,9 +1,12 @@
 import React from "react";
-import "../mocks";
+//import "../mocks";
```
josholdham, line 2: tidy?

packages/ad/__tests__/ios/dom-context.ios.js
```diff
@@ -1,9 +1,12 @@
...
+import { mockNativeModules } from "@times-components-native/mocks";
 import DOMContextNative from "../../src/dom-context";
 import TestRenderer from "react-test-renderer";
 import * as domContextUtils from "../../src/utils/dom-context-utils";
 
+mockNativeModules();
```
josholdham, line 8: is this needed given it's also in the beforeEach?

packages/ad/src/dom-context.tsx
```diff
@@ -178,87 +183,89 @@ const DOMContext = ({
...
+        padding: 0;
+      }
+    </style>
+    <script>
+      window.nuk = {
+        "ads": {
+          "blocked": false,
+          "commercialSection": "comment",
+          "pageTitle": "",
+          "editionDate": "2022-07-12",
```
josholdham, line 205: is this supposed to be there? from the perspective of someone with no knowledge about this (🙃) it looks odd to have these hard-coded/defined editions/dates/articles but they could have no effect or just be some weird quirk...

josholdham (approved):
looks good, just with a few queries similar to before

packages/ad/src/dom-context.tsx
petermason: will be resolved in 1b

packages/ad/__tests__/ios/dom-context.ios.js
petermason: resolved in step 1b

packages/ad/__tests__/ios/dom-context.ios.js
petermason: to resolve in 1b

PR #664, opened 2022-08-17
Feature/tnlt 9632 pass utm values to native module analytics

petermason:
#### Description

If the Topic and Author page is accessed via deeplink return the deeplink parameters back as a tracking event.

#### Checklist
 - [x] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet


packages/pages/src/author-profile/index.js
```diff
@@ -11,7 +11,19 @@ const {
...
+    if (deeplink_value) {
+      track({
+        attrs: {
+          eventTime: new Date(),
+          pageName: "AuthorProfile",
+        },
+        ...deeplink_value,
+      });
+    }
+  }, deeplink_value);
```
josholdham, line 25: Genuine question: does this work correctly when the dependency (deeplink_value) isn't part of an array?

I assume so, but just not seen it before

josholdham (approved):
Good old deeplinks!

I have a couple of comments/questions, but no blockers, so approved ✌️

packages/pages/src/article/index.js
```diff
@@ -24,6 +24,8 @@ const ArticlePage = (props) => {
     onAction: openInBrowser,
   };
 
+  console.log("ARTICLE PROPS: ", props);
```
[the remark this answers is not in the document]
petermason: oops :)

packages/pages/src/author-profile/index.js
petermason: actually, this *should* be in an Array! good catch

packages/pages/src/article/index.js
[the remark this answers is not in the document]
petermason: removed

packages/pages/src/author-profile/index.js
petermason: updated to correct syntax

josholdham:
All looks good - nice one

josholdham (approved):

PR #685, opened 2022-09-15
fix: restore bundle for command ios:build-lib:js

jaglop:
#### Description
fix: restore bundle for command ios:build-lib:js

#### Checklist
 - [ x ] Tested on iOS simulator


josholdham (approved):
nice fix, well found

petermason (approved):

PR #686, opened 2022-09-19
Chore/lint improvement and typescript bump

jaglop:
#### Description
Linting changed files wasn't quickly available so added the option, we might add a pre-commit hook to run `yarn lint:changed` from now on if we agree on it

Also added the option to ensure we adhere to prettier config from the command line with `yarn prettier`, another potential check pre-commit

#### Checklist
 - [ x ] Unit tests


josholdham (approved):
good improvements 🙌

PR #719, opened 2022-10-24
Bugfix(TNLT-9402): inline modal image fix on rotate

josholdham:
#### Description

**About the PR**
I will shortly create a new branch with fixed tests, and create a PR to merge that branch into this one. I have decided to keep it separate so the diff is easier to follow.

I'd suggest reviewing the code diff top to bottom, as some inline comments explaining the rationale behind some changes apply in multiple places, but are summarised once when they are first encountered.

**Code Changes**
In an effort to prevent unnecessary re-renders on article pages:

- Made a number of article components functional rather than class components, so we could use hooks
-- This allowed us to remove the functions from within the context consumers (replacing them with hooks), which should improve performance
-- Functional components are also more modern, and should have performance optimisations under the hood too.

- Separated the frequently used isArticleTablet value from the main responsive context provider, as everywhere it was used was being re-rendered if ANY value rom the provider changed. As the relevant value (isArticleTablet) changed a lot less frequently, putting it in a separate provider prevented a lot of re-renders

- Re-named some article templates, for clarity 

- Anonymous functions inside render blocks can have performance down-sides, so I removed a lot of these too.

#### Checklist
 - [ ] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet


packages/article-comment-tablet/src/article-comment-tablet.js
```diff
@@ -61,127 +72,106 @@ class ArticlePage extends Component {
...
-                </View>
-              </View>
-            )}
-          </Context.Consumer>
-        )}
-      </ResponsiveContext.Consumer>
-    );
-  }
-}
+  return (
```
josholdham, line 125: The diff here looks a lot bigger/worse than it is. It's largely a case of removing the enclosing context consumers - I think the indentation changes have made the diff look crazy. I'll try to comment/highlight any actual changes inline with comments

packages/article-comment-tablet/src/article-comment-tablet.js
```diff
@@ -61,127 +72,106 @@ class ArticlePage extends Component {
...
+          onCommentGuidelinesPress={onCommentGuidelinesPress}
+          onCommentsPress={onCommentsPress}
+          onImagePress={onImagePress}
+          onLinkPress={onLinkPress}
+          onRelatedArticlePress={onRelatedArticlePress}
+          onTooltipPresented={onTooltipPresented}
+          onTopicPress={onTopicPress}
+          onTwitterLinkPress={onTwitterLinkPress}
+          onVideoPress={onVideoPress}
+          onViewableItemsChanged={null}
```
josholdham, line 163: this.onViewableItemsChanged was never defined in the original, so I removed the ternary expression

packages/article-comment-tablet/src/article-comment-tablet.js
```diff
@@ -21,14 +21,15 @@ import {
 } from "./article-prop-types/article-prop-types";
 import styles from "./styles";
 
-class ArticlePage extends Component {
-  constructor(props) {
-    super(props);
-    this.renderHeader = this.renderHeader.bind(this);
-  }
+const ArticleCommentTablet = props => {
```
josholdham, line 24: I renamed some of the article templates to be more descriptive and in line with their file names.

packages/article-image/src/inset-caption.js
```diff
@@ -10,7 +10,7 @@ const captionStyle = {
 };
 
 const CaptionComponentPrimaryNative = ({ text, credits, CaptionComponent }) => {
-  const { isArticleTablet } = useResponsiveContext();
+  const { isArticleTablet } = usePartialResponsiveContext();
```
josholdham, line 13: A lot of article templates and components only relied on `isArticleTablet` from the responsive provider, which rarely changes in the life cycle of an app. 

However, being imported from a context provider which also contains other more frequently changing values, the use of this hook was causing frequent re-renders in the components referencing them.

I made another context provider that only provides this value.

Probably not the neatest solution, but I think other than replacing the entire provider with something like Redux it might be the best way for now....

packages/article-comment-tablet/src/article-comment-tablet.js
```diff
@@ -40,8 +41,18 @@ class ArticlePage extends Component {
       standfirst,
       template,
     } = article;
+
     const showLeadAsset = template === "magazinecomment";
 
+    const renderCaption = ({ caption }) => (
```
josholdham, line 47: Anonymous functions within render properties can have a performance impact, so I tried to hoover some up when I saw them

petermason (approved):

packages/responsive/src/responsive.tsx
```diff
@@ -21,12 +23,23 @@ const ResponsiveProvider: React.FC = ({ children }) => {
     calculateResponsiveContext(width, height, fontScale),
   );
 
+  const [partialState, setPartialState] = useState({
```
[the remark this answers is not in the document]
petermason: This is a great PR and the resulting conversation has been very useful. Nice work

PR #744, opened 2022-11-28
Fix(TNLT-10307): destroy ads on article disappear

josholdham:
#### Description

Previously, the ad slot was being destroyed when the article disappeared event was received, but then immediately reloaded/replaced.

The correct behaviour is the ad slot remains destroyed until/unless the article re-appears again.

You can see in the below video that the ad is reloaded after swiping back to article.

Note: Ovi has also implemented a fix (in the ads' prod release) for a javascript error that occurred during the call to `dynamicRemoveSlotRender`

#### Screenshots 

https://user-images.githubusercontent.com/1517021/204323734-5411466c-0318-4f5b-b53d-e354451b299e.mp4

#### Checklist
 - [ ] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet


josholdham:
> Looks good to me but I reckon the native iOS changes are accidental and shouldn't be leaked into the PR, or am I missing something?

Good question. I rebased from master and ran yarn install, and I think the updates happened during that. I could remove the changes though, as the diff with master probably isn't... good.

petermason (approved):

PR #856, opened 2023-05-05
Fix - add check for perumtive in ad webview

petermason:
#### Description

Add a conditional check for the existence of permutive inside the ad webview, before calling a function against it

#### Checklist
 - [x] Unit tests
 - [ ] Updated E2E tests
 - [ ] Updated storybook
 - [ ] Tested on iOS simulator
 - [ ] Tested on Tablet


skumar-newsuk (approved):
Looks good to me.

